PR #125, opened 2022-10-19
Enable statistics endpoint for mosaics

### LINKED ISSUE #70: Enable mosaic based stats  requests
[The MosaicTilerFactory](https://github.com/stac-utils/titiler-pgstac/blob/b3b4b40c42b2f4009ab4e078a1ee09d3d61f0f3a/titiler/pgstac/factory.py#L417) enables to request statistics for a `searchId` (instead of a single COG url or STAC Item Id). This makes it possible to generate stats for an AOI in a single step (as opposed to retrieving a list of COGs within the AOI and then requesting the stats for each). 

The danger is that API will need load every file within the AOI into memory in order to generate the stats, which can become very slow/memory intensive with a large number of files. The nature of the VEDA dashboard is such that each stats query is always bounded to a single day, so I assume that there will never be more than a few dozen files being opened. 

In order to enable the endpoint, you just have to add: 
```python
add_statistics=True
```
 to the [MultiBaseTilerFactory](https://github.com/developmentseed/eoAPI/blob/c32a205099e5938dd2070e3496de098abec469b8/src/eoapi/raster/eoapi/raster/app.py#L43-L48) instantiation

Again, a big thank you to @vincentsarago for being one step ahead of everyone else and having already implemented the functionality before we even realized we might need it!

ividito:
# What
- Mosaics now support the `/statistics` endpoint - resolves #70.
- Docker-compose now supports hot-reload for changes within their respective `/runtime/src/` folders
- Local Dockerfiles use the same Python version as our hosted lambdas
- `pygeoif` constrained - the new version released in late September breaks other dependencies and prevents the STAC api from running.

# Why
- Local development was a bit cumbersome, with at least one error on start and a chunk of time needed to see changes after making them.
- This makes it possible to generate stats for an AOI in a single step (as opposed to retrieving a list of COGs within the AOI and then requesting the stats for each).

# How tested

Manually tested locally on a handful of mosaics defined in the `workflows` folder.

- [ ] TODO Add a reasonable automated test for statistics endpoint and test on hosted dev stack


ividito:
Got around to trying a datetime filter, as well as further pared-down test data. Still ran into significant performance issues. After discussing with the team, I'm closing this PR and issue for now, as there isn't an immediate need for this functionality.

PR #139, opened 2023-01-04
Collection search function

### LINKED ISSUE #131: Collection search
To fully implement the [analysis page dataset selection](https://github.com/NASA-IMPACT/delta-ui/issues/210) we need a way to know what collections have items for a given aoi and date range.

It would be something like the `/search` endpoint, but instead of returning items, it would just returns a list of collections.

cc @nerik @hanbyul-here @anayeaye

moradology:
This PR adds the ability to search through collections with temporal and spatial boundaries as is available for items

Closes #131


moradology:
Closing as aimed at wrong branch

PR #140, opened 2023-01-04
Collection search

### LINKED ISSUE #131: Collection search

moradology:
This PR adds the ability to search through collections with temporal and spatial boundaries as is available for items.

A new function, `pgstac.collection_id_search` has been added to the DB and new endpoints have been wired up in stac-fastapi at `/collection-id-search`. This endpoint takes intersection, bbox, and datetime queries just as the item search does and appropriate models have been added to ensure correctness of generated documentation. 
image

Closes #131

### DIFF
--- a/database/runtime/handler.py
+++ b/database/runtime/handler.py
@@ -196,21 +196,24 @@ def create_collection_search_functions(cursor) -> None:
 
         RETURN QUERY
         EXECUTE format('SELECT
-            oq.id
+            oq.id as id
         FROM
             (SELECT
                 iq.id as id,
                 ST_MakeEnvelope(
                     (iq.spatial->0)::INTEGER, (iq.spatial->1)::INTEGER, (iq.spatial->2)::INTEGER, (iq.spatial->3)::INTEGER, 4326
                 ) as geometry,
-                iq.temporalmin as temporalmin,
-                iq.temporalmax as temporalmax
+                lower(iq.temporalrange) as temporalmin,
+                upper(iq.temporalrange) as temporalmax
             FROM
                 (SELECT
                     id,
                     (content::jsonb->''extent''->''spatial''->''bbox''->0) as spatial,
-                    (content::jsonb->''extent''->''temporal''->''interval''->0->>0)::timestamptz as temporalmin,
-                    (content::jsonb->''extent''->''temporal''->''interval''->0->>1)::timestamptz as temporalmax
+                    parse_dtrange(
+                        coalesce((content::jsonb->''extent''->''temporal''->''interval''->0->>0), ''..'')
+                        || ''/'' ||
+                        coalesce((content::jsonb->''extent''->''temporal''->''interval''->0->>1), ''..'')
+                    ) as temporalrange
                 FROM collections
                 ) as iq
             ) as oq
--- a/database/runtime/handler.py
+++ b/database/runtime/handler.py
@@ -183,6 +183,60 @@ def create_collection_summaries_functions(cursor) -> None:
     """
     cursor.execute(sql.SQL(periodic_datetime_summary_sql))
 
+    search_collections_sql = """
+    CREATE OR REPLACE FUNCTION collectionsearch(_search jsonb) RETURNS setof text AS $$
+        DECLARE
+            where_segments text[];
+            _where text;
+            geom geometry;
+            dtrange tstzrange;
+            sdate timestamptz;
+            edate timestamptz;
+        BEGIN
+            IF j ? 'datetime' THEN
+                dtrange := parse_dtrange(j->'datetime');
+                sdate := lower(dtrange);
+                edate := upper(dtrange);
+
+                where_segments := where_segments || format(' temporalmin <= %L::timestamptz AND temporalmax >= %L::timestamptz ',
+                    edate,
+                    sdate
+                );
+            END IF;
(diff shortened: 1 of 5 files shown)


database/runtime/handler.py
```diff
@@ -183,6 +183,63 @@ def create_collection_summaries_functions(cursor) -> None:
     """
     cursor.execute(sql.SQL(periodic_datetime_summary_sql))
 
+    search_collections_sql = """
```
anayeaye, line 186: nit: can we put the search collections definition in its own `create_search_collections_functions()` instead of inside of the `create_collection_summaries_functions()` just for future readability? 

Or if we want to fit all the collection-sh function definitions into one long function maybe shift the summary function to the end (instead between the datetime functions) and update the docstrings?
---
Also: have you tried this out in the dev stack? I'm tempted to just add the function manually to try it out--can always drop it after testing and increment the DB SCHEMA version to make sure the pgstac bootstrapper deployment runs the new definition

moradology: Yeah, I've tested the function out and I *believe* it does what is expected. Should work on dev or even locally if you have an instance

anayeaye: Good point, I'll run locally!

stac_api/runtime/src/api.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def register_post_search(self):
+        """Register search endpoint (POST /search).
+        Returns:
+            None
+        """
+        super().register_post_search()
+        self.router.add_api_route(
+            name="Search",
+            path="/collection-search",
```
anayeaye, line 25: @moradology what do you think about `/collection-search-ids` or something that specifies the endpoint will return a list of IDs? 

I suspect we will eventually also want to implement a `collection-search` that returns an array of collections in the same response format as a `/collections` request. By making this list of IDs endpoint unique, we will not have to coordinate a response format change with users in the future.

moradology:
Here's the function in practice (note that the dataset here is not from our project, this is output from directly testing against a pgstac instance with the function added in this PR):
image
image

j08lue:
How to populate local db: https://github.com/NASA-IMPACT/veda-backend/blob/53074600d14fc818244a7a4969af0dece3c44000/.github/workflows/pr.yml#L51-L54

stac_api/runtime/src/core.py
```diff
@@ -0,0 +1,110 @@
...
+                    """,
+                    req=req,
+                )
+                collections = await conn.fetch(q, *p)
+        except InvalidDatetimeFormatError:
+            raise InvalidQueryParameter(
+                f"Datetime parameter {search_request.datetime} is invalid."
+            )
+
+        return [collection["id"] for collection in collections]
```
anayeaye, line 57: I will revisit the manual function definition in my local db, but I am currently getting a result without an id key (instead it is the name of the function). `[<Record collection_id_search='noaa-emergency-response'>]`

We can make that work with `return [collection["collection_id_search"] for collection in collections]` but we could also touch up that function to return an object with an id key or I think we can return an array of strings without a key which would probably have to be unpacked from a tuple result as `return [collection for collection in collections]`.

anayeaye (approved):
Approved, but let's do some testing with veda-config after the dev api updates are deployed.
I bumped the version of `VEDA_DB_SCHEMA_VERSION` stored in the secrets for the dev deployment environment so the github actions should trigger the database schema upgrade in this PR.

I was able to load some test data and perform basic collection id search testing locally (I'll share some test data in a separate issue). 

I also added this snippet to `.github/workflows/tests/test_stac.py` and ran locally to confirm regular item searches are working (I wasn't expecting these changes to impact item searches but we didn't have anything covered).

```
def test_search_api():
    """test item search"""
    # item search
    query = {"collections": ["noaa-emergency-response"], "filter-lang": "cql-json"}
    resp = httpx.post(f"{stac_endpoint}/search", json=query)

    assert resp.status_code == 200
    items = resp.json()["features"]
    assert len(items) == 10

    # item search with should return 0 items
    query = {
        "collections": ["noaa-emergency-response"], 
        "datetime": "2004-03-07T00:00:00Z/2004-03-11T00:00:00Z", 
        "filter-lang": "cql-json"
        }
    resp = httpx.post(f"{stac_endpoint}/search", json=query)

    assert resp.status_code == 200
    items = resp.json()["features"]
    assert len(items) == 0
```

anayeaye:
Bummer. The deploy action failed. Looks like we just need to adjust what role defines the new db function? I am going to make sure there wasn't a previously defined version of function and retrigger the deploy action. 
```
{
    "Status": "FAILED",
    "Reason": "See the details in CloudWatch Log Stream: 2023/02/07/[$LATEST]01ad9239e5b049e19848dc4164a04e35",
<SNIP>
    "Data": {
        "message": "must be owner of function collection_id_search"
    }
}
```

moradology:
😅  I'm afraid I don't fully understand what went wrong here. What role did you end up modifying, out of curiosity?

PR #146, opened 2023-01-17
feat: dynamically inject tiler endpoints into items

moradology:
This PR adds links to WMTS XML and tile previews dynamically. This pattern should enable us to modify advertised links without changing the underlying records in PGSTAC.

# Note
The query parameters anticipated by `titiler` will likely need to be configured on a per-collection basis and there's no obvious sense about how that should work yet. This is desirable as we should try to keep our previews etc. looking the same for the same pieces of data. Should the dashboard pull its configuration from some shared location which STAC API can also access? Should the STAC API retrieve these details by querying the dashboard? This is not yet clear.

Closes https://github.com/NASA-IMPACT/veda-architecture/issues/123


moradology:
Also of note here: this PR is on top of the collection-id-search because it also depends on a custom `CoreCrudClient` implementation.

stac_api/runtime/src/links.py
```diff
@@ -0,0 +1,129 @@
...
+    ) -> None:
+        """Initialize a LinkInjector"""
+        self.collection_id = collection_id
+        # The collection_id should be suitable for getting a RenderConfig with more details
+        self.render_config = get_render_config()
+        self.tiler_href = tiles_settings.titiler_endpoint or ""
+
+    def inject_collection(self, collection: Collection) -> None:
+        """Inject rendering links to a collection"""
+        collection.get("links", []).append(self._get_collection_map_link())
```
slesaad, line 41: 🤔

moradology: Cherry picked this out, so I think this is leftover from the merge. Fixing!

anayeaye:
Some supporting docs for running locally to test: https://github.com/NASA-IMPACT/veda-backend/issues/151. I'm hoping to have a look at this issue but these notes might help someone else run these updates locally sooner

smohiudd:
We now have a way to include render parameters as metadata in collections and items using this new stac extension: https://github.com/stac-extensions/render

Here is an example of what it would look like in geoglam: https://dev.delta-backend.com/api/stac/collections/geoglam

@moradology what is the status of this PR? Can we reconfigure to use the render extension?

moradology:
This PR is set up to dynamically inject URLs to rendering links on titiler (with appropriate query parameters). What it leaves out is support for non-default cases (handled by hidden configs on planetery computer, I believe). So it sounds like the ask here would be to see if rendering definitions are provided by this extension and for each one produce a set of links that titiler makes good on? I think that's doable!

smohiudd:
Along with including the params in the titiler links as you mentioned, another thing we were thinking about was getting a "rendered_preview" href as an item asset link similar to how planetary computer does it: https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-1-rtc/items

stac_api/runtime/src/render.py
```diff
@@ -0,0 +1,117 @@
...
+        return f"&{get_param_str(self.render_params)}"
+
+    class Config:
+        """Pydantic config class for RenderConfig."""
+
+        json_loads = orjson.loads
+        json_dumps = orjson_dumps
+
+
+def get_render_config() -> RenderConfig:
```
smohiudd, line 113: Can we get the render parameters from the `render` extension from the collection?

moradology: Absolutely, that makes sense to me

stac_api/runtime/src/core.py
```diff
@@ -108,3 +111,47 @@ async def collection_id_get_search(
...
+        Returns:
+            ItemCollection containing items which match the search criteria.
+        """
+        _super: CoreCrudClient = super()
+        request = kwargs["request"]
+
+        result = await _super._search_base(search_request, **kwargs)
+        render_params = {}
+
+        if len(result["features"]) > 0:
```
smohiudd, line 141: Can someone have look a this logic and let me know if it makes sense?

anayeaye:
this work got picked up in #247

Issue #147, opened 2023-01-24
Error requesting statistics using a FeatureCollection


ividito:
Originally flagged by @danielfdsilva 

When making a request to to the /statistics endpoint with a Feature Collection, it fails with a 500 error.

url: `https://staging-raster.delta-backend.com/cog/statistics?url=s3://veda-data-store-staging/no2-monthly/OMI_trno2_0.10x0.10_202001_Col3_V4.nc.tif`

body: `{"type":"FeatureCollection","features":[{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-11.25,35.73945],[-10.54687,45.73885],[4.21875,43.82883],[3.86719,33.4873],[-11.25,35.73945]]]}},{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-14.41409,50.35231],[5.62497,49.90155],[-5.62503,60.8073],[-14.41409,50.35231]]]}}]}`

Using single features, or a MultiPolygon instead works fine.

The error appears when a newer (>0.8) version of `titiler` is used - this bug might be a good reason to work through the breaking changes of a core dependency upgrade. There may be front-end changes required as part of this upgrade. Most breaking changes are documented [here](https://github.com/developmentseed/titiler/blob/master/CHANGES.md).

anayeaye:
Note on the test geojson example: geojson lint flags the object as invalid. Here is a fixed geojson. The input adjustment does not change the issue in this ticket but I wanted to share in case anyone else goes down that road.

`{"type": "FeatureCollection", "features": [{"type": "Feature", "properties": {}, "geometry": {"type": "Polygon", "coordinates": [[[-11.25, 35.73945], [3.86719, 33.4873], [4.21875, 43.82883], [-10.54687, 45.73885], [-11.25, 35.73945]]]}}, {"type": "Feature", "properties": {}, "geometry": {"type": "Polygon", "coordinates": [[[-14.41409, 50.35231], [5.62497, 49.90155], [-5.62503, 60.8073], [-14.41409, 50.35231]]]}}]}`

vincentsarago:
The issue is somewhere in pydantic/geojson-pydantic but I'm not able to pinpoint exactly where. I tried to only update those library (not updating rio-tiler) but it didn't solve the issue :-(

PR #148, opened 2023-01-25
Upgrade titiler and titiler-pgstac

ividito:
Fixes: #147 

Tested with a local DB instance, front-end seemed to work fine with a handful of collections. This upgrade might not require as many front-end changes as we expected.

One thing this does break/change are `asset_expressions` ([ref1](https://github.com/developmentseed/titiler/blob/master/CHANGES.md#080-2022-12-01), [ref2](https://github.com/stac-utils/titiler-pgstac/blob/master/CHANGES.md#020-2022-12-13)).


ividito:
Ready for review now that #149 has been merged.

vincentsarago:
by using `titiler-pgstac==0.2.0` you can remove the custom factory in https://github.com/NASA-IMPACT/veda-backend/blob/develop/raster_api/runtime/src/factory.py#L61-L258

which is now directly in titiler-pgstac https://github.com/stac-utils/titiler-pgstac/blob/0.2.0/titiler/pgstac/factory.py#L721-L879 (you need to pass `add_mosaic_list=True` in `MosaicTilerFactory(add_mosaic_list=True)` instead of https://github.com/NASA-IMPACT/veda-backend/blob/develop/raster_api/runtime/src/app.py#L50

.github/workflows/pr.yml
```diff
@@ -30,17 +30,16 @@ jobs:
...
-          AWS_SECRET_ACCESS_KEY=${{secrets.AWS_SECRET_ACCESS_KEY}} \
-          docker-compose up --build -d stac raster
-
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.8'
+          python-version: '3.9'
+
+      - uses: KengoTODA/actions-setup-docker-compose@v1
```
anayeaye, line 38: I'm not that familiar with the subtle differences between `docker-compose` and `docker compose`. Would things work just changing the launch command to `docker compose up --build -d`? 

The docker-compose is deprecated warning on the KengoTODA page is why I ask. If it's a can of worms getting docker compose without the hyphen to work (it does for me locally anyway), let's open an issue to stop needing it in the future.

ividito: Good catch - I had missed the warning, and also wasn't familiar with the difference of `docker-compose` vs `docker compose`

anayeaye:
I rebuilt and ran with docker locally. Looks like `/cog/viewer` is replaced by `/cog/map` which offers more configuration options but may be a tad less intuitive for less technical users. I.e these are the differences I noticed and want to confirm are expected:
- users must use url parameter in browser because there isn't an automatic input (I think most everyone does this anyway)
- underlying map context is no longer present
- mini configuration side bar is gone (although the side bar offered less options than the new /cog/map endpoint provides)

former-cog-viewer

cog-map

If this is the expected behavior I'll approve the changes, we probably need to do a light refactor on our data provider documentation to make sure users know how to the new endpoint works and to update any references to the old viewer.

j08lue:
Hmm, from the new viewer template, it seems like the UI elements for setting some of the URL parameters should still be there: https://github.com/NASA-IMPACT/veda-backend/pull/148/files#diff-42a81a69a9d962db79e4c01b3f757dbdb666fa17d46967b4e030ab8e2a3b1070R1. So the absence from your deployment might not be intended, @anayeaye.

@vincentsarago, what do you think about the expected changes with our update here regarding `/cog/viewer` -> `/cog/map`?

From a project perspective: we have not been relying a lot on `/cog/viewer` or promoted the use of it for our data contributors and curators, but it has come in handy sometimes to discover and debug data issues. So if we can keep a functioning viewer that at least has a background map (so users can confirm correct geo-reference fo their data), that would be great. If not, no problem to remove it.

vincentsarago:
`/cog/viewer` should still be there 
https://github.com/NASA-IMPACT/veda-backend/blob/1ea51c6f6fe8458d2b09be5c0af1c646f52aa0e5/raster_api/runtime/src/app.py#L95-L107

there is also a new `/cog/map` that is provided directly by the tiler factory

anayeaye (approved):

PR #155, opened 2023-02-21
upgrade pypgstac to be consistent with veda-data-pipelines

ranchodeluxe:
Related PR:

https://github.com/NASA-IMPACT/veda-data-pipelines/pull/280

This PR just makes `veda-backend` deps match `veda-data-pipeline` deps for `pypgstac`. 

Read through the `pgstac` CHANGELOG and nothing between versions `0.6.6.` and `0.6.11` stands out as backward incompatible (i mean, it is a patch upgrade). So we should be good with this.

This upgrade is to get around a dependency issue `pypgstac` had using an old and `smart-open` package that poetry had trouble dealing with

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ extra_reqs = {
         "pytest",
         "pytest-asyncio",
         "httpx",
-        "pypgstac==0.6.6",
+        "pypgstac==0.6.11",
         "psycopg[binary, pool]",
     ],
 }


ranchodeluxe:
doh, look at that wonderful CI failure 😮‍💨

ranchodeluxe:
> doh, look at that wonderful CI failure 😮‍💨

working on figuring this out. I feel like patch updates (if following semantic versioning) shouldn't really need to do a whole db update: https://developmentseed.slack.com/archives/C04KHH66Q/p1676987445928639 because we're confident patch changes are backward compatible

Issue #154, opened 2023-02-20
Investigate upgrading pgSTAC version


anayeaye:
# What
The pgSTAC version used by the veda-backend is behind the main release. Investigate the lift to upgrade the version so that it is easier to integrate fixes.

# PI Objective
https://github.com/NASA-IMPACT/veda-architecture/issues/164 

# AC Answer These Questions
- [ ] should we upgrade pgSTAC?
- [ ] what projects will need to be updated if the version is changed (i.e. the veda-stac-ingestor uses a versioned pypgstac)?
- [ ] what other steps would be needed to upgrade and migrate the database to a newer version of pgSTAC?

anayeaye:
Some pros for upgrading to > 0.7.0
- from David: the temporal extent in the code that is in 0.7 takes the full extent from the min(datetime) max(end_datetime)
- There is functionality for updating collection extents automatically (no manual trigger)

smohiudd:
I had a chat with @bitner about this and it seems the pgstac upgrade migration will be straight forward without any breaking changes. It should be able to run in the cdk bootstrap in `veda-backend`:

`Migrate(pgdb).run_migration(params["pgstac_version"])` 

Also the automatic collection extent update can be implemented using the pgstac_settings table. But note that the extent update can't handle multi polygon extents and instead will have a single bounding box extent - I don't think this should be a problem in veda. 

Another thing to note about v0.7.0 is that it includes a "use_queue" configuration parameter that allows deferred processing for building indexes, adding constraints and updating partitions. This would effectively defer these overhead processes for later using some chron job and would overcome having to complete them at time of ingest. This is an opt in feature so not necessary to implement right now, but something to think about for our stack.

smohiudd:
The pgstac upgrade planning ticket is documented here: https://github.com/NASA-IMPACT/veda-architecture/issues/242#issuecomment-1524111402. This one can be closed.

PR #159, opened 2023-03-09
Support optional AWS requester-pays S3 transfer costs configuration

anayeaye:
# What
This PR adds optional raster-api configuration to pay all S3 transfer costs. The setting is applied globally to all requests by setting `AWS_REQUEST_PAYER=requester` in the raster-api lambda environment.

# Why
We want to use data hosted by an external in-region provider that requires that the requester pays S3 transfer costs.

# Assumptions
VEDA data read role is already restricted to only in-region buckets so requester pays costs will be negligible. 

# How tested
Manually deployed the CloudFormation update to the dev stack and confirmed that several raster-api read requests from different buckets were successful.

# Further work
- [ ] Costs should be monitored to see if further restrictions are needed to limit costs (i.e. are we incurring excessive costs requesting data from buckets that do not require the requester pay S3 transfer costs).
- [ ] Tests are now run against a newer version of pgstac than we are actually running, if we do not upgrade in #154, we should update our documentation to clarify why the versions differ

> **Note** To use this optional parameter this change for an operation veda-backend stack:
> 1. Add VEDA_RASTER_AWS_REQUEST_PAYER=requester to the secret containing the environment variables for a given stage
> 2. If using `VEDA_RASTER_DATA_ACCESS_ROLE_ARN`, confirm that the role has a policy specifically allowing GET requests to the in-region requester-pays bucket.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,7 +87,7 @@ services:
 
     depends_on:
       - database
-    command: bash -c "bash /tmp/scripts/wait-for-it.sh -t 25 -h database -p 5432 && /start.sh"
+    command: bash -c "bash /tmp/scripts/wait-for-it.sh -t 120 -h database -p 5432 && /start.sh"
     volumes:
       - ./scripts:/tmp/scripts
       - $HOME/.aws/credentials:/root/.aws/credentials
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -102,7 +102,6 @@ services:
       - POSTGRES_DB=postgis
       - PGUSER=username
       - PGPASSWORD=password
-      - PGHOST=localhost
       - PGDATABASE=postgis
     ports:
       - "5432:5432"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -95,7 +95,7 @@ services:
   database:
     container_name: veda.db
     platform: linux/amd64
-    image: ghcr.io/stac-utils/pgstac:v0.6.6
+    image: ghcr.io/stac-utils/pgstac:v0.7.1
     environment:
       - POSTGRES_USER=username
       - POSTGRES_PASSWORD=password
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     depends_on:
       - database
       - raster
-    command: bash -c "bash /tmp/scripts/wait-for-it.sh -t 45 -h database -p 5432 && /start.sh"
+    command: bash -c "bash /tmp/scripts/wait-for-it.sh -t 120 -h database -p 5432 && /start.sh"
     volumes:
       - ./scripts:/tmp/scripts
 
--- a/raster_api/runtime/setup.py
+++ b/raster_api/runtime/setup.py
@@ -9,6 +9,7 @@ inst_reqs = [
     "titiler.pgstac==0.1.0.a9",
     "titiler.application>=0.5,<0.6",
     "starlette-cramjam>=0.1.0,<0.2",
+    "fastapi==0.93.0",
     "importlib_resources>=1.1.0;python_version<'3.9'",
     "aws_xray_sdk>=2.6.0,<3",
     "aws-lambda-powertools>=1.18.0",
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ extra_reqs = {
         "pytest",
         "pytest-asyncio",
         "httpx",
-        "pypgstac==0.6.6",
+        "pypgstac==0.7.1",
         "psycopg[binary, pool]",
         "pydantic[dotenv]==1.9,<1.10",
     ],
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ extra_reqs = {
         "httpx",
         "pypgstac==0.6.6",
         "psycopg[binary, pool]",
+        "pydantic[dotenv]==1.9,<1.10",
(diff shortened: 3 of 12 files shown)


anayeaye:
Pre-commit is running locally for me but I am on python 3.9 (not 3.8 like the github runner). I'm sorting through the debug messages in the failed lint action.

setup.py
```diff
@@ -19,7 +19,7 @@
         "pytest",
         "pytest-asyncio",
         "httpx",
-        "pypgstac==0.6.6",
+        "pypgstac==0.7.1",
```
vincentsarago, line 22: you may need to update https://github.com/NASA-IMPACT/veda-backend/blob/develop/docker-compose.yml#L98 to 0.7.1

vincentsarago: https://github.com/NASA-IMPACT/veda-backend/pull/159/commits/d3c249309ffe0c3fc26dabdac506813f108b849d 🙈

docker-compose.yml
```diff
@@ -102,7 +102,6 @@ services:
       - POSTGRES_DB=postgis
       - PGUSER=username
       - PGPASSWORD=password
-      - PGHOST=localhost
```
vincentsarago, line 105: https://github.com/stac-utils/pgstac/issues/166#issuecomment-1446525911

vincentsarago: still getting 

```
Run pypgstac pgready --dsn ***0.0.0.0:5432/postgis
[13](https://github.com/NASA-IMPACT/veda-backend/actions/runs/4383620233/jobs/7674104761#step:7:14)
error connecting in 'pool-1': connection failed: server closed the connection unexpectedly
[14](https://github.com/NASA-IMPACT/veda-backend/actions/runs/4383620233/jobs/7674104761#step:7:15)
	This probably means the server terminated abnormally
[15](https://github.com/NASA-IMPACT/veda-backend/actions/runs/4383620233/jobs/7674104761#step:7:16)
	before or while processing the request.
```

cc @bitner

vincentsarago: note: CI won't necessary `fail` if ☝️ step fails so we should add a `test` to make sure this step worked

vincentsarago: fixed: pgstac needed a bit more time to get up 🥳

vincentsarago:
🤷 that `predeploy` step fails because of `plpygis` module which is shipped with the `.pyc` files 🤦 
Screenshot 2023-03-10 at 12 20 08 PM

I've started a PR over https://github.com/bosth/plpygis/pull/8 but it's unlikely that it will get merged/deploy soon

multiple solutions:
- force install of plpygis from my fork
- get rid of plpygis usage in pgstac (cc @bitner)
- remove the whole `.py/.pyc` handling in the docker build https://github.com/NASA-IMPACT/veda-backend/blob/develop/stac_api/runtime/Dockerfile#L9-L13 but I'll still worry about the `.pyc` files being provided in the module

setup.py
anayeaye: I was hoping that the 0.7.1 pypgstac wouldn't have problems with writing to a minor version below but hadn't looked very closely at the container we use for the tests. 

This seems like a good solution for us because we may want to take advantage of some of the other upgrades in 0.7.* pgstac anyway. #154

anayeaye:
Thank you for getting these action unblocked @vincentsarago!

Since the first bullet worked for the pre-deploy, do you think it'll be safe to just use the same for our deploy on merge and then plan on watching for the merge on you plpygis pr and/or removal of plpygis from pgstac?
> multiple solutions:
>  - force install of plpygis from my fork
>  - get rid of plpygis usage in pgstac (cc @bitner)
> - remove the whole .py/.pyc handling in the docker build https://github.com/NASA-IMPACT/veda-backend/blob/develop/stac_api/runtime/Dockerfile#L9-L13 but I'll still worry about the .pyc files being provided in the module

vincentsarago:
🥳 https://github.com/bosth/plpygis/issues/7#issuecomment-1464870708 plpygis should be fixed soon, let's wait for it

vincentsarago:
🤔 tests are now failing because of a pydantic issue in the raster api 🤷 

```
veda.raster  |     result = context.run(func, *args)
veda.raster  |   File "/opt/bitnami/python/lib/python3.9/site-packages/titiler/pgstac/factory.py", line 359, in register_search
veda.raster  |     model.Link(
veda.raster  |   File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
veda.raster  | pydantic.error_wrappers.ValidationError: 1 validation error for Link
```

vincentsarago:
@anayeaye I have no idea about what's going on and I can't reproduce locally :-( 
my hope is that by upgrading titiler-pgstac in #148 this could fix the issue

raster_api/runtime/setup.py
```diff
@@ -9,6 +9,7 @@
     "titiler.pgstac==0.1.0.a9",
     "titiler.application>=0.5,<0.6",
     "starlette-cramjam>=0.1.0,<0.2",
+    "fastapi==0.93.0",
```
vincentsarago, line 12: Comparing logs between 🟢 and 🔴 I saw that the only differences were the FastAPI and Starlette versions. This PR fix FastAPI to the latest `working` version. 

I'll check in titiler-pgstac if something is broken with FastAPI>=0.94

vincentsarago (approved):

Issue #168, opened 2023-03-30
Modify private subnet types in shared VPC


ividito:
When deploying our VPC, we receive the following deprecation warning:

```
[WARNING] aws-cdk-lib.aws_ec2.SubnetType#PRIVATE_WITH_NAT is deprecated.
  use `PRIVATE_WITH_EGRESS`
  This API will be removed in the next major release.
```

We should change our private subnets to avoid errors. In order to make this change safely (without impacting staging), we will need to create a dev VPC to test the changes.

### AC

- [x] Deprecation resolved, private subnet type reviewed
- [x] Dev VPC created for testing
- [x] If needed. unused VPCs removed to make space for dev VPC

PR #173, opened 2023-04-26
RDS in Private Subnet

smohiudd:
Include `VEDA_DB_PUBLICLY_ACCESSIBLE` env variable which determines if the RDS should be provisioned in a private subnet.


database/infrastructure/construct.py
```diff
@@ -127,10 +127,10 @@ def __init__(
         stack_name = Stack.of(self).stack_name
 
         # Configure accessibility
-        publicly_accessible = False if veda_db_settings.private_subnets else True
+        publicly_accessible = True if veda_db_settings.publicly_accessible else False
```
anayeaye, line 130: nit: we don't need to infer `publicly_accessible` from subnet config here if it is already defined. We could directly use use `veda_db_settings.publicly_accessible` on line 172.

smohiudd: So my thinking was to only use the `veda_db_settings.publicly_accessible` config to determine if the rds is in a private subnet. I removed the `veda_db_settings.private_subnets` config. So publicly_accessible isn't inferred right now.

database/infrastructure/config.py
```diff
@@ -34,11 +34,13 @@ class vedaDBSettings(BaseSettings):
             "**Once used always REQUIRED**"
         ),
     )
-    private_subnets: Optional[bool] = Field(
-        False,
-        description="Boolean deploy database to private subnets",
+    # private_subnets: Optional[bool] = Field(
+    #     True,
+    #     description="Boolean deploy database to private subnets",
+    # )
```
smohiudd, line 40: I removed the private_subnets config

database/infrastructure/construct.py
anayeaye: I agree with that logic, what I mean is that on line 176 and 191 you could replace `publicly_accessible=publicly_accessible,` with `publicly_accessible=veda_db_settings.publicly_accessible` to use the configuration directly and delete the local variable `publicly_accessible`. Private subnets are directly based on `veda_db_settings.publicly_accessible` not the local variable on line 133. 

Really using veda_db_settings directly instead of a local variable is just a style preference, the result is the same either way, so either way is works.

smohiudd: Oh ok that makes sense! Thanks for the clarification. I'll make your suggested change.

network/infrastructure/construct.py
```diff
@@ -48,7 +48,7 @@ def __init__(
             )
             private_subnet = aws_ec2.SubnetConfiguration(
                 name="private",
-                subnet_type=aws_ec2.SubnetType.PRIVATE_WITH_NAT,
+                subnet_type=aws_ec2.SubnetType.PRIVATE_ISOLATED,
```
ividito, line 51: This change also closes #168

ividito (approved):

PR #178, opened 2023-05-17
[WIP] Release changes

anayeaye:
# Changes
- `/collection-id-search` endpoint searches for all items meeting filter criteria and returns unique ids
- Titiler and titiler-pgstac upgrades
- Python upgrade to 3.9
- RDS public accessibility now configurable


anayeaye:
https://github.com/NASA-IMPACT/veda-ui/pull/534

anayeaye:
https://github.com/NASA-IMPACT/veda-backend/issues/179

database/runtime/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM lambci/lambda:build-python3.8
+FROM public.ecr.aws/sam/build-python3.9:latest
```
slesaad, line 1:
```suggestion
FROM --platform=linux/amd64 public.ecr.aws/sam/build-python3.9:latest
```

slesaad: i've had issues with deploying from the new mac, this solves it

slesaad: wait, not just mac, but also from github actions 🤔

anayeaye: Even with `export DOCKER_DEFAULT_PLATFORM=linux/amd64`? This has been deploying from github actions as is. I'm reading up, though, to see if we should take out a branch from develop to make the change.

smohiudd: Isn't github actions running on ubuntu so shouldn't need `--platform=linux/amd64`?

smohiudd (approved):

database/runtime/Dockerfile
slesaad: thought so too, maybe i am misremembering. but it's always a good idea to add it whereever.

ividito (approved):

raster_api/runtime/src/app.py
```diff
@@ -61,20 +71,24 @@
...
+    environment_dependency=settings.get_gdal_config,
     router=APIRouter(route_class=LoggerRouteHandler),
 )
 app.include_router(stac.router, tags=["Items"], prefix="/stac")
 
 cog = TilerFactory(
     router_prefix="/cog",
     optional_headers=optional_headers,
-    gdal_config=settings.get_gdal_config(),
+    environment_dependency=settings.get_gdal_config,
```
ranchodeluxe, line 82: I don't see us changing `ApiSettings.get_gdal_config` to be a property or anything here so I think we're not calling the function and hence not passing the auth info

ranchodeluxe: Isayah is saying this gets called downstream by `Depends`

smohiudd: So `environment_dependency` is a callable in titiler. That was one of the breaking changes when `gdal_config` was removed.

PR #174, opened 2023-05-10
Add ingest API and bulk ingestor to veda-backend

ividito:
Issue: #172 

This PR adds the `ingest-api` service, which includes the bulk loader and certain ingestion endpoints currently administered by `veda-stac-ingestor`. The goal is for `veda-stac-ingestor` to be an entrypoint for our data pipelines, while this service handles interactions with `pgstac`.

[Diff ](https://github.com/NASA-IMPACT/veda-stac-ingestor/compare/dev...dnm/ingest-api) against `veda-stac-ingestor` -> this will make it easier to see actual changes from our existing ingest-api service, which is largely ported functionality from `veda-stac-ingestor`.

Other meaningful changes to call out:
- Added a Dynamo serialization change from `cdk-pgstac`, which let me remove our old serializer hacks for dealing with large numbers.
- Added CDK to add `ingest-api` to our stack, and tried to set it up in a way that config values are imported from existing resources in the stack (whereas in `veda-stac-ingestor`, they would have been required env vars)

Minor changes I threw in:
- github action jobs are renamed, which lets us test them individually using `act -j job-name` (before this change, this would run all 3 variations of each job)
- Our pattern for adding OIDC support and improving env management is ported over, but not fully implemented in our actions and docs yet

Issue #172, opened 2023-04-25
Create Ingest API Service in veda-backend


ividito:
## Background

Currently, `veda-backend` contains database, STAC, and raster tiling services, as well as standalone VPC infrastructure. Another repo, `veda-stac-ingestor`, contains an API for interacting with the ingestion process, both through direct insertion to a database and by submitting jobs to a pipeline that then interacts with the database.

We want to create a pattern in veda-backend that is comparable to `cdk-pgstac`, which also has an ingestor-api as its own module. This new service will centralize all database interactions within `veda-backend`, improving maintainability while also allowing the addition of any valid STAC object to the pgstac DB without necessarily requiring the object to contain VEDA/dashboard-specific rules defined in veda-stac-ingestor.

We also want to apply a DynamoDB serialization change that was [implemented ](https://github.com/developmentseed/cdk-pgstac/pull/26)in `cdk-pgstac` - it makes sense to include this as part of this transition.

As a note - we aren't implementing `cdk-pgstac` within VEDA at this point, but the structure of that project aligns with our goals, and following that project's patterns could make it easier to transition to `cdk-pgstac` in the future.

## AC

- [x] Create a new ingest API service using a structure consistent with the current STAC and raster API services.
- [x] Add infrastructure to manage our DynamoDB batch loading system
- [x] Migrate all database and DynamoDB interactions (insertion, update, delete) from `veda-stac-ingestor` to the new ingest API service. Interactions with the ingestion pipeline are excluded from this migration.
- [x] Apply DynamoDB serialization change to the new ingest API service.
- [x] Using the new ingest API service, support adding any valid STAC object to the pgstac DB.
- [x] Where possible, migrate unit tests from `veda-stac-ingestor`, and add them to `veda-backend` (the testing structure is somewhat different between the two repsitories)

### Desired state (from https://github.com/NASA-IMPACT/veda-architecture/issues/215)
![image](https://user-images.githubusercontent.com/19193384/234296402-360d5a1b-766e-4977-bb5c-5d3e50a32e44.png)

ividito:
Current state - there is a PR (#174)  up with functionality ported over from `veda-stac-ingestor`. Based on feedback, there's a draft PR(#191) open with some CDK restructuring to make our constructs more similar to what we have for our existing services, in order to maintain consistency in our repo.

[...]

PR #191, opened 2023-06-27
Ingest construct refactor

ividito:
Related: #172 
Based on PR: #174 

Splitting ingest API stack into re-usable constructs - this is closer to what `cdk-pgstac` does, except in Python and with some of our veda-specific functionality included. There are also some other fixes and improvements to the `feature/ingest-api` branch included here.

The `cdk diff` for this branch is equivalent to `feature/ingest-api`, except for the added domain for the ingest API and the updated lambda assets.


anayeaye:
superseded by #258

PR #202, opened 2023-07-25
Bug/nat sg inbound cidr

ranchodeluxe:


smohiudd:
I don't know for sure if CDK will smartly connect everything the right order. Wondering if @ividito or @anayeaye have any thoughts?

ividito:
This definitely will not synthesize, taking a look now to see what the workaround would be (probably allowing default SGs to be created, then adding rules to it after the VPC is declared)

smohiudd:
Doesn't look like SG can be created without a VPC: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_ec2/SecurityGroup.html

ividito:
The NAT provider will create a default one when the VPC is created, we can grab it and add rules to it after that. This order of operations synthesizes, and as far as I can tell the resulting graph looks good.

```
        nat_provider_instance = aws_ec2.NatProvider.instance(
            instance_type=aws_ec2.InstanceType("t3.nano"),
        )

        vpc = aws_ec2.Vpc(
            self,
            "vpc",
            max_azs=base_settings.vpc_max_azs,
            cidr=base_settings.vpc_cidr,
            subnet_configuration=[public_subnet, private_subnet],
            nat_gateway_provider=nat_provider_instance,
            nat_gateways=base_settings.vpc_nat_gateways,
        )

        nat_sg = nat_provider_instance.security_group

        # Allow all outbound traffic
        nat_sg.add_egress_rule(
            aws_ec2.Peer.any_ipv4(),
            aws_ec2.Port.all_traffic(),
            "Allow all outbound traffic"
        )

        # Allow inbound traffic from the VPC's CIDR
        nat_sg.add_ingress_rule(
            aws_ec2.Peer.ipv4(vpc.vpc_cidr_block),
            aws_ec2.Port.all_traffic(),
            "Allow inbound traffic from the VPC's CIDR block"
        )
```

smohiudd (approved):

ividito (approved):

PR #205, opened 2023-07-26
Fix invalid string in rule description

anayeaye:
Tiny changes to make the stand alone base vpc construct deployable


ividito (approved):

ranchodeluxe (approved):
doh

PR #213, opened 2023-08-23
Use NAT Gateway Service

ranchodeluxe:
### Background

https://github.com/NASA-IMPACT/veda-analytics/issues/86

### Next Steps
- [x] deploy with a separate `base_name` and make sure the network is setup correctly. Maybe even manually provision a Lambda in it to do the public test that is failing on GHGC
- [ ]  test this change against the [veda-shared-base/network/vpc in UAH](https://us-west-2.console.aws.amazon.com/vpc/home?region=us-west-2#VpcDetails:VpcId=vpc-0512162c42da5e645)
- [ ] Then promote it to GHGC downstream fork and test there

### DIFF
--- a/standalone_base_infrastructure/network_construct.py
+++ b/standalone_base_infrastructure/network_construct.py
@@ -34,6 +34,7 @@ class BaseVpcConstruct(Construct):
             max_azs=base_settings.vpc_max_azs,
             cidr=base_settings.vpc_cidr,
             subnet_configuration=[public_subnet, private_subnet],
+            nat_gateways=base_settings.vpc_nat_gateways,
         )
 
         vpc_endpoints = {
--- a/standalone_base_infrastructure/network_construct.py
+++ b/standalone_base_infrastructure/network_construct.py
@@ -25,7 +25,7 @@ class BaseVpcConstruct(Construct):
             name="private",
             # NOTE: this line automatically creates a NAT Gateway for each AZ
             # and binds the route table in the private subnet
-            subnet_type=aws_ec2.SubnetType.PRIVATE_WITH_NAT,
+            subnet_type=aws_ec2.SubnetType.PRIVATE_WITH_EGRESS,
         )
 
         vpc = aws_ec2.Vpc(
--- a/standalone_base_infrastructure/network_construct.py
+++ b/standalone_base_infrastructure/network_construct.py
@@ -25,7 +25,7 @@ class BaseVpcConstruct(Construct):
             name="private",
             # NOTE: this line automatically creates a NAT Gateway for each AZ
             # and binds the route table in the private subnet
-            subnet_type=ec2.SubnetType.PRIVATE_WITH_NAT,
+            subnet_type=aws_ec2.SubnetType.PRIVATE_WITH_NAT,
         )
 
         vpc = aws_ec2.Vpc(
--- a/standalone_base_infrastructure/network_construct.py
+++ b/standalone_base_infrastructure/network_construct.py
@@ -23,11 +23,9 @@ class BaseVpcConstruct(Construct):
         )
         private_subnet = aws_ec2.SubnetConfiguration(
             name="private",
-            subnet_type=aws_ec2.SubnetType.PRIVATE_ISOLATED,
-        )
-
-        nat_provider_instance = aws_ec2.NatProvider.instance(
-            instance_type=aws_ec2.InstanceType("t3.nano"),
+            # NOTE: this line automatically creates a NAT Gateway for each AZ
+            # and binds the route table in the private subnet
+            subnet_type=ec2.SubnetType.PRIVATE_WITH_NAT,
         )
 
         vpc = aws_ec2.Vpc(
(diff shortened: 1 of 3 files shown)


ranchodeluxe:
Question: will CDK do a force replacement here of the whole VPC or will do an in-place edit of the existing VPC. Hopefully the latter.

ividito:
> Question: will CDK do a force replacement here of the whole VPC or will do an in-place edit of the existing VPC? Hopefully the latter.

`cdk diff` makes me think that the VPC will change in-place, but the route tables will be recreated (which should be fine)

standalone_base_infrastructure/network_construct.py
```diff
@@ -23,11 +23,9 @@ def __init__(
...
         private_subnet = aws_ec2.SubnetConfiguration(
             name="private",
-            subnet_type=aws_ec2.SubnetType.PRIVATE_ISOLATED,
-        )
-
-        nat_provider_instance = aws_ec2.NatProvider.instance(
-            instance_type=aws_ec2.InstanceType("t3.nano"),
+            # NOTE: this line automatically creates a NAT Gateway for each AZ
+            # and binds the route table in the private subnet
+            subnet_type=aws_ec2.SubnetType.PRIVATE_WITH_NAT,
```
ividito, line 28: I get a new warning here: 
```
[WARNING] aws-cdk-lib.aws_ec2.SubnetType#PRIVATE_WITH_NAT is deprecated.
  use `PRIVATE_WITH_EGRESS`
  This API will be removed in the next major release.
  ```

**Review by ividito — COMMENTED** [review]
Mostly LGTM, but I'm not 100% sure what the reasoning was for some of the code being removed (is there a reason we would ever need more than one NAT gateway?).

File: standalone_base_infrastructure/network_construct.py (hunk 1 above)
Comment (anayeaye) [reply standalone_base_infrastructure/network_construct.py:28]: @ividito I've been wondering about whether we should be switching to PRIVATE_WITH_EGRESS as well. Maybe we can see what changes this PR causes and test the subnet type change separately.

Comment (ranchodeluxe) [reply standalone_base_infrastructure/network_construct.py:28]: Let me make that change and redeploy on my test instance

**Comment by ranchodeluxe** [comment]
> Mostly LGTM, but I'm not 100% sure what the reasoning was for some of the code being removed (is there a reason we would ever need more than one NAT gateway?).

Are you referring to the number of NAT(s) per AZ here? Or is this a general question about removal of EC2 Instances that function as a NAT and instead using the NAT Gateway?

**Comment by ranchodeluxe** [comment]
Here is my deployed [test VPC](https://us-west-2.console.aws.amazon.com/vpc/home?region=us-west-2#VpcDetails:VpcId=vpc-0d688dbb235b98ad3) and[ test Lambda](https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/functions/test-timeout?newFunction=true&tab=code) showing everything is gravy

Before we'd do this we kinda need to know if CDK is going to do a whole teardown here b/c that means the site will be offline and might need to redeploy all services. Let's talk about it at standup

File: standalone_base_infrastructure/network_construct.py (hunk 1 above)
Comment (ranchodeluxe) [reply standalone_base_infrastructure/network_construct.py]: Change committed

**Comment by anayeaye** [comment]
_Before_ the addition of the `PRIVATE_WITH_EGRESS` change I did some sanity checking:
- I updated a non-customer facing shared vpc stack that already had a veda-backend instance deployed. The change was made in place (vpc endpoints/route tables were updated and the nat Instance was replaced by a nat gateway)
- I confirmed basic functionality of the veda-backend instance running in the updated shared vpc after the update was completed

I'll pull the change and run these sanity checks again

**Comment by anayeaye** [comment]
No changes caused by changing the subnet type in the construct 👍

**Comment by ranchodeluxe** [comment]
> > Mostly LGTM, but I'm not 100% sure what the reasoning was for some of the code being removed (is there a reason we would ever need more than one NAT gateway?).
> 
> Are you referring to the number of NAT(s) per AZ here? Or is this a general question about removal of EC2 Instances that function as a NAT and instead using the NAT Gateway?

I actually think @ividito (of course he is!). I know everyone loves CDK here but there's too much magic. The default behavior here is that CDK magically creates a NAT Gateway for each AZ. I looked at my Terraform for all VPCs I use and we can easily just use one for both private subnets. But that means I have manually do all the plumbing.

Thoughts @ividito or @anayeaye? Should we just do that?

**Review by anayeaye — APPROVED** [review]

**Comment by ividito** [comment]
> Are you referring to the number of NAT(s) per AZ here? Or is this a general question about removal of EC2 Instances that function as a NAT and instead using the NAT Gateway?

I was talking about removing the config parameter + related code: `vpc_nat_gateways: Optional[int] = 1`. I wasn't sure if this was still configurable behavior we need to support. The new way seems more appropriate, but I know we sometimes have unique networking requirements in different accounts.

**Comment by anayeaye** [comment]
Sorry I didn't notice there was a still a discussion going when I approved. RE multiple nat gateway count--it is still configurable via aws-cdk and it looks like it [determines the number of gateways per availability zone](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.Vpc.html#natgateways). I suppose leaving the configuration in place with default=1 gives us that option later?

If we're still on a roll, we could address this, too, I suppose
```
WARNING] aws-cdk-lib.aws_ec2.VpcProps#cidr is deprecated.
  Use ipAddresses instead
```

ranchodeluxe:
@ividito, @anayeaye: my most recent commits make everything groovy:

1. I added back in `nat_gateways=base_settings.vpc_nat_gateways,` and it did what I wanted and make both private subnets point to the NAT Gateway (only one this time)

2. I couldn't find the right incantation to fix this warning: `[WARNING] aws-cdk-lib.aws_ec2.VpcProps#cidr is deprecated.
  Use ipAddresses instead` but we can do that later

PR #222, opened 2023-09-08
Add STAC browser from eoapi-cdk constructs

emileten:
Closes https://github.com/NASA-IMPACT/veda-architecture/issues/305. 

This PR adds a STAC browser deployment to the VEDA backend stack, [using the eoapi-cdk constructs](https://github.com/developmentseed/eoapi-cdk). 

The following code changes were required : 

- Exposing the STAC API URL from the construct class so that I have access to it when instantiating the `StacBrowser` eoapi-cdk construct (we need the URL for the deployment).
- Added the latest version of eoapi-cdk to the `deploy` python dependencies. ⚠️ ⚠️ I had to bump `aws-cdk.aws_apigatewayv2_integrations_alpha` and `aws-cdk.aws_apigatewayv2_alpha` to 2.47.0.a0 as a consequence, because of `eoapi-cdk` requirements. Is this going to break anything here ? ⚠️ ⚠️ 
- lastly, instantiating the `StacBrowser` construct in `app.py`. Notes : (1) I am using the non-custom STAC URL (2) for now, not using any cloudfront distribution, just simple static website hosting and therefore (3) made the resulting bucket public.


app.py
```diff
@@ -69,6 +70,15 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
     domain_name=domain.stac_domain_name,
 )
 
+stac_browser = StacBrowser(
+    veda_stack,
+    "stac-browser",
+    github_repo_tag="v3.1.0", # hard coded to the latest for now. 
+    stac_catalog_url=stac_api.stac_api_url, # using the non-custom-domain for now.
```
smohiudd, line 77: I think we might need to add the `api/stac` path prefix here

emileten: Added it directly in the stac-api `construct`

emileten:
I guess the `predeploy` workflow fails because I don't have the required settings in the fork ?

emileten:
So unfortunately I had to change the plan. To build the browser one needs to know what the STAC API URL is, and it's determined only _after_ deployment of veda-backend. So, with a dynamic URL determined by the CDK, the only way is to deploy the browser _after_ veda-backend, in a separate stack. I changed things following this approach, adding a `stac_browser` folder for its own small stack deployment.

That being said, we don't have this problem if we have a stable STAC API URL, which I believe is the plan ? Like, we already have "https://dev-stac.delta-backend.com/api/stac/" -- I am not sure where that's deployed at the moment though. I see this `alt_stac_api` in `app.py` but that doesn't seem to be used at the moment in existing deployments. Maybe in `veda-routes`? 

So, for now, I just took out the stac browser deployment, requiring that the person deploying provides a STAC_API_URL environment variable at deployment. 

We can change that later and reintegrate the browser to the veda-backend stack.

emileten:
Also, note that I have a running test deployment for this in the VEDA account @smohiudd

anayeaye:
@emileten we do have to a stable stac api url info available before deployment--I am going to try and make time to try out some minor changes off of your branch this week to show what I am thinking.

emileten:
@anayeaye yes I saw that e.g. https://dev-stac.delta-backend.com/api/stac/ is online. Ideally, I would have preserved the first version of this PR, which had the browser deployed at the same time as the rest in `veda-backend/app.py` with a STAC API URL pulled from settings provided by the user. 

However, I wasn't comfortable doing this because I am confused as to _where_ that URL is actually defined across the VEDA ecosystem : 

1.  [This in `veda-backend`](https://github.com/NASA-IMPACT/veda-backend/blob/b9dceef367b7ce73afeaba5f190cfb0de008cd30/app.py#L73) makes me think that URL is deployed here. If it was the case, then we obviously _shouldn't_ also have that URL as a setting for the browser here. 
2.  However, if I look at the actual `veda-backend` deployments on the AWS console, I do not see this URL being a part of the stack, so it makes me think it's deployed somewhere else. `veda-routes`, maybe ? It's not in the code though. 

#### TL;DR, I guess my question is : in the future, where is the stable STAC API URL going to be deployed ? If outside of `veda-backend` (something like 2), then I can revert my last change.

smohiudd:
@emileten the STAC api uses API gateway and is output here:

https://github.com/NASA-IMPACT/veda-backend/blob/e783a5acfb7fe8bc22df41291c679b18fdc9d113/stac_api/infrastructure/construct.py#L96C9-L96C56

As you know there will also be a cloudfront distribution but that shouldn't matter when it comes the STAC api you need for the stac browser.

anayeaye:
@emileten + @smohiudd sorry for the extremely slow communications this week. Here's what I was thinking would work for importing and using the eoapi-cdk browser in the main app.py. The dev stack you are looking at is deployed automatically from a github action that looks up all the configuration in AWS Secrets Manager. I think your work would fit into the mainline by adding a couple new configuration parameters. This is a messy example of what the new config might look like and how it could be used to generate a new bucket and deploy a browser with or without a known cloudfront url (because we're still working on that part).


**`config.py`**
New configuration in config.py. We currently use environmental AWS Secrets Manager secrets stored for each stage. For the dev deployment (secrets stored in a secret named `veda-backend-uah-dev-env` we could add a new `VEDA_CLOUDFRONT_STAC_URL` variable when we have our cloudfront setup. When this isn't configured, we could continue to use the API url from the `stac-api` construct.
```
veda_browser_bucket_arn: Optional[str] = Field(
    None,
    description="Arn of existing s3 static webiste bucket. Necessary when deploying to higher security accounts.",
)

veda_cloudfront_stac_url: Optional[str] = Field(
    None,
    description="Externally managed cloud front STAC API url",
)
veda_stac_path_prefix: Optional[str] = Field(
    "",
    description="Optional path prefix to add to all api endpoints",
)
```

**`app.py`**
To deploy the imported browser with the other veda-backend constructs. Some of the conditional flow is clunky but I think it is a start? 
```
<SNIP>
# create a new bucket if not provided. this is something we'd usually want in it's own infrastructure/ directory 
# but maybe we could start with this and then work to make the bucket construct importable down the road
if not veda_app_settings.browser_bucket_arn:
    stac_browser_bucket = s3.Bucket(
        browser_stack,
        "stac-browser-bucket",
        bucket_name=f"{APP_NAME}-{STAGE_NAME}",
        removal_policy=RemovalPolicy.DESTROY,
        auto_delete_objects=True,
        website_index_document="index.html",
        public_read_access=True,
        block_public_access=s3.BlockPublicAccess(
            block_public_acls=False,
            block_public_policy=False,
            ignore_public_acls=False,
            restrict_public_buckets=False,
        ),
        object_ownership=s3.ObjectOwnership.OBJECT_WRITER,
    )
    browser_bucket_arn = stac_browser_bucket.bucket_arn
else:
    browser_bucket_arn = veda_app_settings.browser_bucket_arn


# use a cloufront url in stac browser if provided
if stac_api_url := veda_app_settings.cloudfront_stac_url:
    if veda_app_settings.stac_path_prefix:
        stac_api_url = f"{veda_app_settings.cloudfront_stac_url}/{veda_app_settings.stac_path_prefix}"
    else:   
        stac_api_url = veda_app_settings.cloudfront_stac_url

stac_browser = StacBrowser(
    veda_stack,
    "stac-browser",
    github_repo_tag="v3.1.0",  # hard coded to the latest for now.
    stac_catalog_url=stac_api_url,  
    bucket_arn=browser_bucket_arn,
)
```

emileten:
@anayeaye thanks for your input. Can you take a look at the current changes ? I reverted the creation of the separate stack and changed a little bit the initial approach. Here's is what's happening : 

in `app.py`, [if a STAC custom domain name was deployed](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/app.py#L76), we first create a [publicly accessible browser](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/app.py#L78). In principle we don't want to have this in `app.py`, but it's temporary until we can use CloufdFront (in which case we will just use either the bucket provided by the user, or the [private bucket created in eoapi-cdk](https://github.com/developmentseed/eoapi-cdk/blob/a1081a22e034d50f89d248d36844d598d45d5c69/lib/stac-browser/index.ts#L25)).
- then we [instantiate](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/app.py#L97) the browser construct. We use as a STAC API url what we find from the `domain` construct instantiated [above](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/app.py#L55). For this to work, I had [to expose a new field](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/domain/infrastructure/construct.py#L112) from the construct. Again, I can't just use the [CfnOutput](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/domain/infrastructure/construct.py#L120) or directly the [domain object](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/domain/infrastructure/construct.py#L91), those won't point me to the actual URL strings when the stack is just synthesized. 

Two of your suggestions that I haven't implemented (yet) : 

- `veda_browser_bucket_arn` in `config.py` : we can add this option when we'll be able to use cloudfront. Right now we're just serving from an S3 bucket and so need to have it accessible. I believe that we typically would use that option for non-public buckets. 
- `veda_cloudfront_stac_url` : we can add this once the URL is _actually_ going to be externally managed. Right now it's not, [since it's deployed here.](https://github.com/emileten/veda-backend/blob/a02be7ec0cccbe673b14f185498ea29f80ce54ea/domain/infrastructure/construct.py#L91).

✅ current changes tested in deployment

anayeaye:
this work was incorporated in #257

PR #225, opened 2023-09-12
fix: handle semantic version releases with tags only

anayeaye:
## What
Remove version management in code (use git tags and releases only). We will have a full changelog in the git release history visible to all users but we will not copy that history to a CHANGELOG.md.

~### NOTE/REQUEST FOR HELP:
Conventional commits are used to infer the next semantic version. The release changelog, however, includes the full commit history. I think this should be something we can configure in the [pyproject commit parser options](https://github.com/NASA-IMPACT/veda-backend/blob/fix/semantic-releases-without-commits/pyproject.toml#L20-L23) but I haven't figured it out yet. [python-semantic-release commit parser config](https://python-semantic-release.readthedocs.io/en/latest/commit-parsing.html).~

## Why
Python-semantic-release commit and changelog updates require an admin [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens) for protected branches (which we do and should have). Using admin personal access tokens is not a sustainable pattern. In the future we may want to consider using a github app to dispense limited permissions tokens for the github action but this also introduces owner-only admin requirements that are difficult for maintainers.

[python-semantic-release requires personal access token for protected branches](https://python-semantic-release.readthedocs.io/en/latest/automatic-releases/github-actions.html#)
> The GITHUB_TOKEN secret is automatically configured by GitHub, with the same permissions as the user who triggered the workflow run. This causes a problem if your default branch is protected.
You can work around this by storing an administrator’s Personal Access Token as a separate secret and using that instead of GITHUB_TOKEN. In this case, you will also need to pass the new token to actions/checkout (as the token input) in order to gain push access.

[commitizen-tools semantic release also requires a personal access token for releases](https://commitizen-tools.github.io/commitizen/tutorials/github_actions/#create-a-new-release-with-github-actions) 
> If you use GITHUB_TOKEN instead of PERSONAL_ACCESS_TOKEN, the job won't trigger another workflow. It's like using [skip ci] in other CI's.

## How tested
I created a temporary branch with a github release action on push to the branch to test out the configuration and github actions changes.

> **Note** I created an init v0.0.0 release manually to avoid creating a first release changelog that described every PR over the last 1.5 years.


ranchodeluxe (approved):
LGTM

ividito (approved):

gadomski (approved):

PR #229, opened 2023-09-25
feat: add configuration for deploying a cloudfront distribution for the veda-backend

smohiudd:
Include Cloudfront Distribution construct in veda-back which should help with the [STAC Browser construct](https://github.com/NASA-IMPACT/veda-backend/pull/222) feature. 

One way to address the API Gateway url not resolving was to use the `restapi_id` in the url pattern: `https://{restapi_id}.execute-api.{region}.amazonaws.com/` to access both the stac and raster apis. 

This branch has been test deployed as the `path-prefix-veda-backend` stack using the `delta-backend.xyz` domain.


routes/infrastructure/construct.py
```diff
@@ -0,0 +1,95 @@
...
+
+        stack_name = Stack.of(self).stack_name
+
+        s3Bucket = s3.Bucket.from_bucket_name(
+            self,
+            "stac-browser-bucket",
+            bucket_name=veda_route_settings.stac_browser_bucket,
+        )
+
+        no_cache_policy = cf.CachePolicy(
```
emileten, line 38: Can we use the `CachingDisabled` [managed policy](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html#managed-cache-policy-caching-disabled) instead ? Looks pretty similar ?

Longer note : 

So we're turning off any caching Cloud Front is doing with that policy. We've actually been doing that on MAAP as well (albeit manually on the console, so I don't have any code to link), after noticing that STAC metadata wasn't getting updated frequently enough on the STAC browser.

However, I'm still puzzled that I am doing this because, isn't the point of CloudFront to cache the content in the closest edge point to the user and decrease latency thanks to that ? If we turn that off entirely, why are we using CloudFront ? 

In the first place I remember that we started to use CloudFront because that's what the [AWS docs tell us to do](https://docs.aws.amazon.com/AmazonS3/latest/userguide/WebsiteHosting.html) for serving static content stored in S3 over HTTPs. However, is the CloudFront layer really needed for this, can't we just create a route53 record that directly points to the bucket ? 

_Note : those are mainly questions, I am not suggesting we stop using CloudFront_

routes/infrastructure/construct.py
```diff
@@ -0,0 +1,95 @@
...
+        # Certificate must be in zone us-east-1
+        domain_cert = (
+            certificatemanager.Certificate.from_certificate_arn(
+                self, "domainCert", veda_route_settings.cert_arn
+            )
+            if veda_route_settings.cert_arn
+            else None
+        )
+
+        self.distribution = cf.Distribution(
```
emileten, line 55: Am I right that with this set of behaviors, the mapping is going to be, assuming the domain is `dev-stac.delta-backend.com` :  

-  `https://dev-stac.delta-backend.com` -> browser
- `https://dev-stac.delta-backend.com/api/stac` STAC API
- `https://dev-stac.delta-backend.com/api/raster` raster API

app.py
```diff
@@ -69,6 +70,14 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
     domain_name=domain.stac_domain_name,
 )
 
+veda_routes = CloudfrontDistributionConstruct(
+    veda_stack,
+    "routes",
+    raster_api_id=raster_api.raster_api.api_id,
+    stac_api_id=stac_api.stac_api.api_id,
```
emileten, line 77: 😱 very nice trick. Spent so much time finding a workaround for this.

emileten (approved):
Looks good to me, do have an important broader question regarding our usage of CloudFront though

ividito (commented):
Going to spend a bit more time looking at this today (particularly looking at how the ingest API is working with the CF distribution in this stack). The failing action will pass if the environment secret is updated with the `ingest_url` and `stac_browser_bucket` values.

smohiudd:
@ividito Also the routes construct isn't configured as as optional (similar to how the domain construct works). So the `ingest_url` and `stac_browser_bucket` configs need to be optional.

routes/infrastructure/config.py
```diff
@@ -0,0 +1,52 @@
...
+    domain_hosted_zone_id: Optional[str] = Field(
+        None, description="Domain ID for the cloudfront distribution"
+    )
+
+    cert_arn: Optional[str] = Field(
+        None,
+        description="Certificate’s ARN",
+    )
+
+    using_mcp_acct: Optional[bool] = False
```
anayeaye, line 42: I think  `using_mcp_acct` is not used? Also, not sure about the intent but possibly we could be inferring this info from whether or not `VEDA_PERMISSIONS_BOUNDARY_POLICY_NAME`  is configured?

anayeaye (commented):
just a couple comments on the config

routes/infrastructure/config.py
```diff
@@ -0,0 +1,52 @@
+"""Settings for Cloudfront distribution - any environment variables starting with
```
anayeaye, line 1: Can you add a line to the readme [advanced config](https://github.com/NASA-IMPACT/veda-backend#advanced-configuration) about this? (sorry if this is a repeat comment)

slesaad:
>One way to address the API Gateway url not resolving was to use the restapi_id in the url pattern: https://{restapi_id}.execute-api.{region}.amazonaws.com/ to access both the stac and raster apis.

I'm kinda surprised that this is the change that made it work. We have the same setup in ghgc, but it still doesn't resolve the urls 🤔 

image
http://staging.ghg.center/api/stac :(

routes/infrastructure/construct.py
```diff
@@ -0,0 +1,96 @@
...
+                bucket_name=veda_route_settings.stac_browser_bucket,
+            )
+
+            no_cache_policy = cf.CachePolicy(
+                self,
+                "no-cache-policy",
+                default_ttl=Duration.seconds(0),
+                min_ttl=Duration.seconds(0),
+                max_ttl=Duration.seconds(0),
+            )
```
slesaad, line 45: there's already no cache policy defined in cdk,

```python
cf.CachePolicy.CACHING_DISABLED
```

smohiudd:
@slesaad I did some experimentation last week and realized that for the cloudfront stac api to work you need the trailing slash... http://staging.ghg.center/api/stac/ 

Without the trailing slash you get 307 temporary redirect the goes to api gateway url

slesaad:
@smohiudd yeah, but the URLs in the catalogs are still the same api gateway urls

smohiudd:
@slesaad the issue with that is cloudfront is not able to forward the `host` header to api gateway. I tried another way which seems to work using a load balancer in this PR: https://github.com/NASA-IMPACT/veda-backend/pull/232

smohiudd:
Thanks @vincentsarago! I got the host header working in APIgateway using the parameter mapping.

routes/infrastructure/construct.py
smohiudd: Done

routes/infrastructure/config.py
smohiudd: Done

stac_api/infrastructure/config.py
```diff
@@ -13,21 +13,26 @@ class vedaSTACSettings(BaseSettings):
...
+        None,
+        description="Domain name for the cloudfront distribution",
+    )
+
     class Config:
         """model config"""
 
         env_file = ".env"
-        env_prefix = "VEDA_STAC_"
+        env_prefix = "VEDA_"
```
anayeaye, line 35: I think this is a breaking change because it causes existing implementations to update the secrets stored in the secrets manager? I think we should address changing `<SCOPE>_VARNAME` across all constructs if we want to break the pattern (especially because it could be useful to share a variable that should be the same in the scopes of different constructs).

Is the change needed now, though, or can we keep scoping our environment variables?

stac_api/infrastructure/construct.py
```diff
@@ -71,12 +71,23 @@ def __init__(
...
+                    aws_apigatewayv2_alpha.ParameterMapping().overwrite_header(
+                        "host",
+                        aws_apigatewayv2_alpha.MappingValue(
+                            veda_stac_settings.domain_hosted_zone_name
+                        )
+                        if veda_stac_settings.domain_hosted_zone_name
+                        else None,
+                    )
+                ),
             )
```
anayeaye, line 91: this is awesome + I wonder if the prefix can be managed here also (in the future)?

Question: I need to try this out but to me it seems that if domain_hosted_zone_name is None, then "host" will be overwritten with None? Or does it just end up being a no-op?

stac_api/infrastructure/config.py
anayeaye: looks like staging has no `VEDA_STAC_` vars in the secrets store so this comment can be downgraded to a nit

stac_api/infrastructure/construct.py
smohiudd: I wanted to make it like a no-op if hosted zone wasn't present. I can double check if that's what it's doing

routes/infrastructure/config.py
```diff
@@ -0,0 +1,50 @@
...
+
+class vedaRouteSettings(BaseSettings):
+    """Veda Route settings"""
+
+    cloudfront: Optional[bool] = Field(
+        False,
+        description="Boolean if Cloudfront Distribution should be deployed",
+    )
+
+    # STAC S#3 browser bucket name
```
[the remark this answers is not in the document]
smohiudd: 1. Yes, for now it's the name of an existing bucket. But that may change depending on how the stac browser construct @emileten is working on is set up
2. Yes you're right they should `None`
3. Agreed

vincentsarago:
@smohiudd Glad it's working 🙏 , sorry for not seeing this issue earlier 😭

routes/infrastructure/construct.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if veda_route_settings.cloudfront:
+            s3Bucket = s3.Bucket.from_bucket_name(
+                self,
+                "stac-browser-bucket",
+                bucket_name=veda_route_settings.stac_browser_bucket,
+            )
+
+            # Certificate must be in zone us-east-1
+            domain_cert = (
```
smohiudd, line 40: I believe the certificate used in cloudfront must be in the us-east-1 region. Right now the arn is being passed manually. Should we be creating a new certificate in us-east-1 instead.

stac_api/infrastructure/construct.py
```diff
@@ -71,12 +71,23 @@ def __init__(
...
+                parameter_mapping=(
+                    aws_apigatewayv2_alpha.ParameterMapping().overwrite_header(
+                        "host",
+                        aws_apigatewayv2_alpha.MappingValue(
+                            veda_stac_settings.domain_hosted_zone_name
+                        )
+                        if veda_stac_settings.domain_hosted_zone_name
+                        else None,
+                    )
+                ),
```
slesaad, line 90: @smohiudd I realized we'd want to do this to the raster API too, because otherwise when we use `/mosaic/register`, it gives back the API gateway URL

smohiudd: Thanks @slesaad have you tested that on GHGC yet?

slesaad: yes! - https://github.com/US-GHG-Center/ghgc-backend/pull/46/files
deployed to dev.ghg.center/api/raster

raster_api/infrastructure/construct.py
```diff
@@ -62,26 +62,37 @@ def __init__(
...
             )
 
         raster_api_integration = (
             aws_apigatewayv2_integrations_alpha.HttpLambdaIntegration(
-                construct_id, veda_raster_function
+                construct_id,
+                handler=veda_raster_function,
+                parameter_mapping=(
+                    aws_apigatewayv2_alpha.ParameterMapping().overwrite_header(
+                        "host",
```
anayeaye, line 88: At least temporarily, I think we need to make this assignment conditional based on whether or not cloudfront is used. It appears that the staging APIs will also be overwritten. Same for the stac api
```
[~] AWS::ApiGatewayV2::Integration raster-api/veda-backend-uah-staging-raster-api/DefaultRoute/raster-api rasterapivedabackenduahstagingrasterapiDefaultRouterasterapi20F21B19 
 └─ [+] RequestParameters
     └─ {"overwrite:header.host":"delta-backend.com"}
```

anayeaye (changes requested):

raster_api/infrastructure/construct.py
smohiudd: I added a conditional for the parameter mapping

anayeaye (approved):

vincentsarago:
@smohiudd @anayeaye https://github.com/developmentseed/titiler/pull/716 might be of interest. Basically when using an API Gateway prefix, it seems that you'll need to set a `root_path` in mangum but also in the fastAPI application as well https://github.com/developmentseed/titiler/blob/main/src/titiler/application/titiler/application/main.py#L65

PR #232, opened 2023-10-05
feat: add load balancer and configuration for deploying cloudfront distribution

smohiudd:
To overcome the issue of the cloudfront base url not being passed to API gateway, this PR instead uses an application load balancer (which was inspired by the work done in CSDA). It also fixes the trailing slash redirect issue we were having with API gateway.

This has been deployed to the `path-prefix` stack and currently working at: [https://delta-backend.xyz/](https://delta-backend.xyz/)

Some things to note:
- the load balancer is not using a certificate so is listening over port 80 (HTTP). SSL is instead being implemented in Cloudfront. I'm not sure if this is a vulnerability. 
- I removed the stac/raster specific route 53 domain configuration 
- In order to pass the host header to the lambda, we need to implement a cloundfront cache policy in the STAC endpoint


vincentsarago:
@smohiudd I'm not sure if the issue is what I think, but when deploying API Gateway with custom hostname, I usually need update the `integration` in API Gateway to overwrite the `hostname` the FastAPI application thinks it is.

Screen Shot 2021-03-09 at 10 46 58 PM

slesaad:
@vincentsarago 🤯 [THAT WORKED](https://dev.ghg.center/api/stac/)! I'd been struggling with this issue for a while, thanks so much!!!!

slesaad:
@smohiudd  i implemented it in ghg center, hope it helps!

```python
stac_api_integration = aws_apigatewayv2_integrations_alpha.HttpLambdaIntegration(
            construct_id,
            handler=lambda_function,
            parameter_mapping=aws_apigatewayv2_alpha.ParameterMapping().overwrite_header(
                "host",
                aws_apigatewayv2_alpha.MappingValue.custom(veda_stac_settings.host),
            )
            if veda_stac_settings.host
            else None,
        )
```

smohiudd:
@slesaad I did the same just now in my old PR! https://github.com/NASA-IMPACT/veda-backend/pull/229
It works great

PR #235, opened 2023-10-10
docs: rename veda-data-pipelines to veda-data-airflow

gadomski:
Updated the diagram too.


gadomski:
Updated w/ **veda-data**.

anayeaye (approved):

PR #237, opened 2023-10-12
fix: include stage subdomain in cloudfront distribution

smohiudd:
The cloudfront distribution includes the `stage` env as the subdomain:

`{stage}.{domain_hosted_zone_name}.com`

### DIFF
--- a/raster_api/infrastructure/construct.py
+++ b/raster_api/infrastructure/construct.py
@@ -24,6 +24,7 @@ class RasterApiLambdaConstruct(Construct):
         self,
         scope: Construct,
         construct_id: str,
+        stage: str,
         vpc,
         database,
         code_dir: str = "./",
@@ -89,7 +90,7 @@ class RasterApiLambdaConstruct(Construct):
             ] = aws_apigatewayv2_alpha.ParameterMapping().overwrite_header(
                 "host",
                 aws_apigatewayv2_alpha.MappingValue(
-                    veda_raster_settings.domain_hosted_zone_name
+                    f"{stage}.{veda_raster_settings.domain_hosted_zone_name}"
                 ),
             )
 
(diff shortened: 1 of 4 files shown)


routes/infrastructure/construct.py
```diff
@@ -53,36 +55,57 @@ def __init__(
...
                     ),
                     cache_policy=cf.CachePolicy.CACHING_DISABLED,
                 ),
                 certificate=domain_cert,
-                domain_names=[veda_route_settings.domain_hosted_zone_name]
+                domain_names=[f"{stage}.{veda_route_settings.domain_hosted_zone_name}"]
                 if veda_route_settings.domain_hosted_zone_name
                 else None,
                 additional_behaviors={
                     "/api/stac*": cf.BehaviorOptions(
```
anayeaye, line 67: I missed this additional behaviors in the last PR, maybe we can drop a TODO here to make this configurable in the condition that a path prefex is used like `f"{veda_route_settings.stac_path_prefix}*": cf.BehaviorOptions(...`. I think we can work on that after using things for a while though, to see what our needs are

raster_api/infrastructure/construct.py
```diff
@@ -89,7 +90,7 @@ def __init__(
             ] = aws_apigatewayv2_alpha.ParameterMapping().overwrite_header(
                 "host",
                 aws_apigatewayv2_alpha.MappingValue(
-                    veda_raster_settings.domain_hosted_zone_name
+                    f"{stage}.{veda_raster_settings.domain_hosted_zone_name}"
```
anayeaye, line 93: comment for future work: maybe we can add a method in the api config files to add a stage subdomain for all non-prod stages? We don't need that now so it's just something to think about.

anayeaye (approved):
I added some discussion comments but no changes requested.

PR #241, opened 2023-10-24
fix(api)!: use root path for prefixing

anayeaye:
# What
- Use fast-api root path configuration and mangum api_base_path instead of overwriting the api host in the http lambda integration
  - Use FastAPI `root_path` for optional base path configuration ~and, if provided, also set proxy url in [additional servers property](https://fastapi.tiangolo.com/advanced/behind-a-proxy/#additional-servers)~
  - Set api_gateway_base_path=root_path in handler

- Simplify custom domain management
  - If using cloudfront, do not create custom api-specific subdomains--just use the default apigateway urls route based on root path settings provided in raster and stac api configuration.

## How tested
Temporary stacks are running to test the two supported path configurations:
1. Using an in-account custom subdomain (`test-raster`) for the raster API: https://test-raster.delta-backend.com/docs
2. Using a cloudfront and custom root path (`/api/raster`) for the raster API: https://alex.delta-backend.xyz/api/raster/docs


smohiudd (approved):

stac_api/runtime/handler.py
```diff
@@ -4,12 +4,15 @@
...
+from src.config import ApiSettings
 from src.monitoring import logger, metrics, tracer
 
+settings = ApiSettings()
+
 logging.getLogger("mangum.lifespan").setLevel(logging.ERROR)
 logging.getLogger("mangum.http").setLevel(logging.ERROR)
 
-handler = Mangum(app, lifespan="auto")
+handler = Mangum(app, lifespan="auto", api_gateway_base_path=settings.root_path)
```
vincentsarago, line 15: instead of using `settings` I think you can use `app.root_path` 

ref https://github.com/developmentseed/eoapi-cdk/pull/82/files#diff-62c863ae9edf662fc81920ac522a18c5650fb0a927fb36b0077e079c4999c88dR9

anayeaye: @smohiudd @ividito: after making 👆 suggested change, I was also able to remove the additional fast api additional servers configuration (sharing because was a red flag in reviews yesterday)

ividito (approved):

PR #245, opened 2023-11-08
fix(apis)!: overwrite api host when provided in config

anayeaye:
# What
- When a custom host is configured, overwrite the host in the api integration (bring back the parameter mapping method introduced in #229)
- Include origin_path in cloudfront ingest api behavior. Note: this path will not work until the root_path is changed to `api/ingest` in the ingest api
- Domain construct now only generates custom subdomains for endpoints when configured (existing stacks must provide new configuration variable). _Note: indentation changes that wrap all subdomain constructs make the change appear larger than they really are. The change removes an unused shared subdomain construct and only generates api-specific subdomains when create subdomains configuration is true._

### New
`VEDA_CUSTOM_DOMAIN` optional configuration to provide a complete url for host overrides in lambda APIs (i.e. `dev.somedomain.com`)
### Removed
`VEDA_DOMAIN_SHARED_SUBDOMAIN` is no longer used (logic is managed with custom subdomains config below)
### New+Breaking
`VEDA_DOMAIN_CREATE_CUSTOM_SUBDOMAINS` boolean defaults false. Existing deployments with custom subdomains for each raster and stac api endpoints need to set this variable to TRUE.


smohiudd:
For the domain construct can we remove:
```
            if veda_domain_settings.shared_subdomain:
                self.shared_sub_domain_name = aws_apigatewayv2_alpha.DomainName(
                    self,
                    "cfSharedApiCustomDomain",
                    domain_name=f"{stage}.{hosted_zone_name}",
                    certificate=certificate,
                )

                CfnOutput(
                    self,
                    "shared-sub-domain-name",
                    value=f"https://{stage}.{hosted_zone_name}/",
                )
```
`self.shared_sub_domain_name` isn't used in any of the other constructs.

smohiudd (approved):

ividito (approved):

PR #248, opened 2023-11-15
fix(raster): temp downgrade rio tiler version

anayeaye:
Temporary rollback rio-tiler to an earlier version


vincentsarago:
ref: https://github.com/cogeotiff/rio-tiler/issues/654

vincentsarago (approved):

PR #250, opened 2023-11-16
fix(raster): pin rio-tiler 6.2.3.post1

anayeaye:
# What
Pin a post rio-tiler release that supports zonal statistics upgrades while a map zooming issue is investigated
# Why
https://github.com/cogeotiff/rio-tiler/issues/654


vincentsarago:
@anayeaye we can now revert to `6.2.7` 🥳 

https://github.com/cogeotiff/rio-tiler/blob/main/CHANGES.md#627-2023-11-29

PR #252, opened 2023-11-21
feat: include app context config for bootstrap qualifier

smohiudd:
Include app context config for bootstrap qualifier.

### DIFF
--- a/config.py
+++ b/config.py
@@ -47,6 +47,11 @@ class vedaAppSettings(BaseSettings):
         description="Custom domain name, i.e. veda-backend.xyz",
     )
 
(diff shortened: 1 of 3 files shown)


config.py
```diff
@@ -47,6 +47,11 @@ class vedaAppSettings(BaseSettings):
         description="Custom domain name, i.e. veda-backend.xyz",
     )
 
+    veda_bootstrap_qualifier: Optional[str] = Field(
+        None,
+        description="Qualifier for cdk bootstraps",
```
anayeaye, line 52: Custom bootstrap qualifier override if not using a default installation of AWS CDK Toolkit to synthesize app.

anayeaye (approved):

PR #257, opened 2023-11-29
feat(browser): use eoapi-cdk to build a radiant earth stac browser

anayeaye:
wip 

> **Note:**
based on work introduced in pr #222 (with updates for current veda-backend)

### DIFF
--- a/app.py
+++ b/app.py
@@ -12,6 +12,7 @@ from permissions_boundary.infrastructure.construct import PermissionsBoundaryAsp
 from raster_api.infrastructure.construct import RasterApiLambdaConstruct
 from routes.infrastructure.construct import CloudfrontDistributionConstruct
 from stac_api.infrastructure.construct import StacApiLambdaConstruct
+from eoapi_cdk import StacBrowser
 
 app = App()
 if veda_app_settings.bootstrap_qualifier:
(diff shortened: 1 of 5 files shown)


app.py
```diff
@@ -89,6 +91,16 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
     region=veda_app_settings.cdk_default_region,
 )
 
+# eoapi-cdk stac-browser only supported for stacks with cloudfront distribution
```
emileten, line 94: Actually you can deploy it without a cloud front distribution !

routes/infrastructure/construct.py
```diff
@@ -32,11 +33,30 @@ def __init__(
...
-            )
+
+            if veda_route_settings.stac_browser_bucket:
+                self.bucket = s3.Bucket.from_bucket_name(
+                    self,
+                    "stac-browser-bucket",
+                    bucket_name=veda_route_settings.stac_browser_bucket,
+                )
+            else:
+                self.bucket = s3.Bucket(
```
emileten, line 44: do we need to make the bucket public now that we're using cloudfront?

emileten (commented):
Question Alex. Here we deploy the browser, taking the cloud front distribution as given. Where is the cloud front distribution CDK code ?

anayeaye:
@emileten re: Where is the cloud front distribution CDK code ?
The cloudfront distribution code is in the `routes/infrastructure` construct.

ividito (approved):

PR #258, opened 2023-11-30
feat(ingest): move ingest API into veda-backend

### LINKED ISSUE #172: Create Ingest API Service in veda-backend

ividito:
Revives and closes #172 

## Why

While working through our production goals (https://github.com/NASA-IMPACT/veda-architecture/issues/332), we found that managing multiple repositories for our backend services wasn't doing us many favors for maintainability. A relevant [ADR ](https://github.com/NASA-IMPACT/veda-architecture/pull/351/files)was made for `veda-features-api`, and many of the arguments for consolidating repositories also applies to `veda-stac-ingestor`

## What

Since #174, we've done significant work refactoring this repo, and have added several infrastructure changes that needed to be accommodated, including new routing mechanisms and Cloudfront support. To maintain consistency, I also had to refactor parts of the ingest API to align with how our other services are provisioned.

## How to review

With very few exceptions (configuration changes + root path support), the runtime code has not changed from `veda-stac-ingestor`. I would suggest mostly [ignoring ](https://github.com/NASA-IMPACT/veda-backend/pull/258/files/d664660cff44d66c689ee5314b47e324bf13a028..4c6e4b4e8b36e886be3abb9315652355fd6e6b3a)this commit when reviewing - it shrinks the number of files in the PR in half.

### DIFF
--- a/app.py
+++ b/app.py
@@ -4,7 +4,7 @@
 import json
 import subprocess
 
-from aws_cdk import App, Stack, Tags, aws_iam
+from aws_cdk import App, Aspects, Stack, Tags, aws_iam
 from aws_cdk import aws_secretsmanager as secretsmanager
 from aws_cdk import aws_ssm as ssm
 from constructs import Construct
--- a/app.py
+++ b/app.py
@@ -13,8 +13,8 @@ from config import veda_app_settings
 from database.infrastructure.construct import RdsConstruct
 from domain.infrastructure.construct import DomainConstruct
 from ingest_api.infrastructure.config import IngestorConfig as ingest_config
-from ingest_api.infrastructure.constructs import ApiConstruct as ingest_api_construct
-from ingest_api.infrastructure.constructs import IngestorConstruct as ingestor_construct
+from ingest_api.infrastructure.construct import ApiConstruct as ingest_api_construct
+from ingest_api.infrastructure.construct import IngestorConstruct as ingestor_construct
 from network.infrastructure.construct import VpcConstruct
 from permissions_boundary.infrastructure.construct import PermissionsBoundaryAspect
 from raster_api.infrastructure.construct import RasterApiLambdaConstruct
@@ -213,6 +213,7 @@ ingest_api = ingest_api_construct(
     config=ingestor_config,
     db_secret=database.pgstac.secret,
     db_vpc=vpc.vpc,
+    domain=domain,
 )
 
 ingestor = ingestor_construct(
@@ -223,6 +224,13 @@ ingestor = ingestor_construct(
     db_secret=database.pgstac.secret,
     db_vpc=vpc.vpc,
 )
+
+veda_routes.add_ingest_behavior(
+    ingest_api=ingest_api.api, stage=veda_app_settings.stage_name()
+)
+
+# Must be done after all CF behaviors exist
+veda_routes.create_route_records(stage=veda_app_settings.stage_name())
 
 
 def register_ssm_parameter(
--- a/app.py
+++ b/app.py
@@ -1,12 +1,20 @@
 #!/usr/bin/env python3
 """ CDK Configuration for the veda-backend stack."""
 
-from aws_cdk import App, Aspects, Stack, Tags, aws_iam
+import json
+import subprocess
+
+from aws_cdk import App, Stack, Tags, aws_iam
+from aws_cdk import aws_secretsmanager as secretsmanager
+from aws_cdk import aws_ssm as ssm
 from constructs import Construct
 
 from config import veda_app_settings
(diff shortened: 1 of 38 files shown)


database/infrastructure/construct.py
```diff
@@ -196,6 +196,8 @@ def __init__(
             database = aws_rds.DatabaseInstance(self, **database_config)
 
         hostname = database.instance_endpoint.hostname
+        self.db_security_group = database.connections.security_groups[0]
```
anayeaye, line 199: we don't need to make the change in this pr but I think at this point we should add a `self.database=database` so that the rds construct in it's entirety is available to other constructs in the deployment (there are a bunch of other database properties that we have added to the construct over time, too)

anayeaye (commented):
I left a couple comments about extra code that I don't think we need in app.py but otherwise I'm excited to just start working with the rest of it.

app.py
```diff
@@ -108,6 +178,60 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+ingestor_config = ingest_config(
+    stac_db_vpc_id=vpc.vpc.vpc_id,
+    stac_db_secret_name=db_secret_name,
+    stac_db_security_group_id=db_security_group.security_group_id,
+    stac_db_public_subnet=database.is_publicly_accessible,
+    stac_api_url=stac_api.stac_api.url,
+    raster_api_url=raster_api.raster_api.url,
+)
+
+env_secret = veda_stack.build_env_secret(
```
anayeaye, line 194: The environment configuration parameters should not be updated by CDK deploy. The build env secret method would be a super useful utility function in support_scripts to facilitate setting up a secret for a new instance.

raster_api/infrastructure/config.py
```diff
@@ -54,7 +54,7 @@ class vedaRasterSettings(BaseSettings):
         description="Name or ARN of the AWS Secret containing database connection parameters",
     )
 
-    raster_data_access_role_arn: Optional[str] = Field(
+    data_access_role_arn: Optional[str] = Field(
```
anayeaye, line 57: the `RASTER` part of the `VEDA_RASTER_` environment prefix was recently stripped (https://github.com/NASA-IMPACT/veda-backend/blob/develop/raster_api/infrastructure/config.py#L86). We may want to bring that back but for now we need that `raster` in the config field name.

ividito: I'm not sure I follow - this change is just so that both the ingest and raster APIs can access the same environment variable. Is there another reason we need to keep `raster_` here?

anayeaye: The current env configuration in the secrets manager for our deployed instances is `VEDA_RASTER_DATA_ACCESS_ROLE_ARN`. We can modify that if needed but we'd also need to make sure we manage that change for all our instances including ghgc. 

Might be simpler to have the ingest assume a raster data access role (which is technically the reason it is used in ingest anyway).

anayeaye (approved):

PR #271, opened 2024-01-11
fix(ingest): use defaults for auth env configuration

anayeaye:
# What
Minor fixes in ingest API to support cognito clients not configured with secrets. Tested end to end token generation, validation, and collection publication in externally deployed instance.

### DIFF
--- a/ingest_api/runtime/src/schemas.py
+++ b/ingest_api/runtime/src/schemas.py
@@ -8,13 +8,7 @@ from typing import TYPE_CHECKING, Dict, List, Optional
 from urllib.parse import urlparse
 
 import src.validators as validators
-from pydantic import (
-    BaseModel,
-    Field,
-    PositiveInt,
-    error_wrappers,
-    validator,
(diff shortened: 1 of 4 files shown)


ingest_api/runtime/src/schemas.py
```diff
@@ -102,9 +90,9 @@ class AuthResponse(BaseModel):
 
 class WhoAmIResponse(BaseModel):
     sub: str = Field(..., description="A unique identifier for the user")
-    cognito_groups: List[str] = Field(
-        ..., description="A list of Cognito groups the user belongs to"
-    )
+    # cognito_groups: List[str] = Field(
+    #     ..., description="A list of Cognito groups the user belongs to"
+    # )
```
ividito, line 95: We can pare this model down even more and allow the auth response from Cognito to add optional fields if needed using `model_config = ConfigDict(extra='allow')`

ividito (approved):

PR #275, opened 2024-01-22
fix: add debugging to authenticate_and_get_token()

botanical:
What:
- Added TimeOut check for debugging purposes

Why:
- dev `/token` endpoint times out


botanical:
Does not actually seem possible to add this so I will close it 😅

Issue #157, opened 2023-03-08
[WIP] Update raster API for alternate assets extension


anayeaye:
# What
Update raster/titiler API to accommodate items that use the alternate assets extension (a common standard used in NASA EarthdataCloud datasets).

# Notes
**extension docs:** [stac-extension/alternate-assets](https://github.com/stac-extensions/alternate-assets)

**Alternate assets examples**
https://landsatlook.usgs.gov/stac-server/collections/landsat-c2l2-sr/items/LE

Detail of a single asset with a login protected https href and an alternate s3 url
<img width="1019" alt="alternate-assets-detail" src="https://user-images.githubusercontent.com/3386265/223802189-c0b7d3f1-c346-46c0-a25e-13c1fad21728.png">
07_L2SR_005113_20010116_20200917_02_T2_SR

**template/explanation as code:**
Here are some notes from @vincentsarago about how we can fetch nested alternate asset urls by building a  TilerFactory using a custom STACReader
```python
import attr
from rio_tiler.io import stac
from rio_tiler.types import AssetInfo
from rio_tiler.errors import InvalidAssetName

@attr.s
class STACReader(stac.STACReader)

    @minzoom.default
    def _minzoom(self):
        return self.tms.minzoom

    @maxzoom.default
    def _maxzoom(self):
        return self.tms.maxzoom
    
    def _get_asset_info(self, asset: str) -> AssetInfo:
        """Validate asset names and return asset's url.
        Args:
            asset (str): STAC asset name.
        Returns:
            str: STAC asset href.
        """
        if asset not in self.assets:
            raise InvalidAssetName(f"{asset} is not valid")

        asset_info = self.item.assets[asset]
        info = AssetInfo(url=asset_info.extra_fields["alternate"]["s3"]["href"])

        if "file:header_size" in asset_info.extra_fields:
            h = asset_info.extra_fields["file:header_size"]
            info["env"] = {"GDAL_INGESTED_BYTES_AT_OPEN": h}

        return info
```

# AC
- [ ] raster-api stac and mosaic endpoints will render data for selected items with standard asset hrefs and items that use the alternate assets extension

moradology:
OK, so I've got this issue pretty well cornered, but there are a couple questions that remain.

1. First one of the *non-trivial* properties on the alt asset from the example is 'requester-pays'. Any ideas on this question? https://github.com/developmentseed/titiler/discussions/758

[...]

PR #282, opened 2024-01-24
feat: alternate asset support in titiler

moradology:
This PR adds support for alternate assets.

It is not as general as it could be, so there's a question here about what we aim to support. Non-s3 backends for these alt-assets are not currently supported. If we want that, we *could* add to the path arguments or else query params expected. Thoughts?

### DIFF
--- a/raster_api/runtime/src/alternate_reader.py
+++ b/raster_api/runtime/src/alternate_reader.py
@@ -1,0 +1,55 @@
+"""Custom Alternate Asset Reader."""
+
+import attr
+from rio_tiler.errors import InvalidAssetName, MissingAssets
+from rio_tiler.types import AssetInfo
+
+from titiler.pgstac.reader import PgSTACReader
+
+
+@attr.s
+class PgSTACReaderAlt(PgSTACReader):
+    """Custom STAC Reader for the alternate asset format used widely by NASA.
+
+    Only accept `pystac.Item` as input (while rio_tiler.io.STACReader accepts url or pystac.Item)
+
+    """
+
+    def _get_asset_info(self, asset: str) -> AssetInfo:
+        """Validate asset names and return asset's url.
+        Args:
+            asset (str): STAC asset name.
+        Returns:
+            str: STAC asset href.
+        """
+        if asset not in self.assets:
+            raise InvalidAssetName(f"{asset} is not valid")
+
+        asset_info = self.input.assets[asset]
+        extras = asset_info.extra_fields
+
+        if ("alternate" not in extras) or ("s3" not in extras["alternate"]):
+            raise MissingAssets("No alternate asset found")
+
+        info = AssetInfo(url=extras["alternate"]["s3"]["href"], metadata=extras)
+
+        info["env"] = {}
+
+        if "file:header_size" in asset_info.extra_fields:
+            h = asset_info.extra_fields["file:header_size"]
+            info["env"].update({"GDAL_INGESTED_BYTES_AT_OPEN": h})
+
+        if requester_pays := extras["alternate"]["s3"].get("storage:requester_pays"):
+            if requester_pays:
+                info["env"].update({"AWS_REQUEST_PAYER": "requester"})
+
+        if bands := extras.get("raster:bands"):
+            stats = [
+                (b["statistics"]["minimum"], b["statistics"]["maximum"])
+                for b in bands
+                if {"minimum", "maximum"}.issubset(b.get("statistics", {}))
+            ]
+            if len(stats) == len(bands):
+                info["dataset_statistics"] = stats
+
+        return info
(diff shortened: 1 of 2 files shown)


moradology:
@vincentsarago This is slightly different than the notes here https://github.com/NASA-IMPACT/veda-backend/issues/157
Where I've done things differently, it has been in trying to keep closer to the reader implementation on titiler-pgstac. Does this seem like a reasonable departure?

raster_api/runtime/src/alternate_reader.py
```diff
@@ -0,0 +1,96 @@
...
+
+        info = AssetInfo(url=extras["alternate"]["s3"]["href"], metadata=extras)
+
+        if "file:header_size" in asset_info.extra_fields:
+            h = asset_info.extra_fields["file:header_size"]
+            info["env"] = {"GDAL_INGESTED_BYTES_AT_OPEN": h}
+
+        if requester_pays := extras["alternate"]["s3"].get("storage:requester_pays"):
+            if requester_pays:
+                info["env"] = {"AWS_REQUEST_PAYER": "requester"}
```
vincentsarago, line 85: maybe there is no `file` extension but just in case I think we should not overwrite the `info["env"]` completely 

```python
info["env"] = {}

if "file:header_size" in asset_info.extra_fields:
    h = asset_info.extra_fields["file:header_size"]
    info["env"].update({"GDAL_INGESTED_BYTES_AT_OPEN": h})

if requester_pays := extras["alternate"]["s3"].get("storage:requester_pays"):
    if requester_pays:
        info["env"].update({"AWS_REQUEST_PAYER": "requester"})
```

raster_api/runtime/src/alternate_reader.py
```diff
@@ -0,0 +1,96 @@
...
+from rasterio.crs import CRS
+from rio_tiler.constants import WEB_MERCATOR_TMS, WGS84_CRS
+from rio_tiler.errors import InvalidAssetName, MissingAssets
+from rio_tiler.io import BaseReader, MultiBaseReader, Reader
+from rio_tiler.io.stac import DEFAULT_VALID_TYPE, _get_assets
+from rio_tiler.types import AssetInfo
+
+
+@attr.s
+class PgSTACReaderAlt(MultiBaseReader):
```
vincentsarago, line 18: Q: why do we need to copy most of the code from titiler-pgstac? 

can't we just do 
```python
from titiler.pgstac.reader import PgSTACReader

@attr.s
class PgSTACReaderAlt(PgSTACReader):

    def _get_asset_info(self, asset: str) -> AssetInfo:
        """Validate asset names and return asset's url.
        Args:
            asset (str): STAC asset name.
        Returns:
            str: STAC asset href.
        """
        if asset not in self.assets:
            raise InvalidAssetName(f"{asset} is not valid")

        asset_info = self.input.assets[asset]
        extras = asset_info.extra_fields

        if ("alternate" not in extras) or ("s3" not in extras["alternate"]):
            raise MissingAssets("No alternate asset found")

        info = AssetInfo(url=extras["alternate"]["s3"]["href"], metadata=extras)

        if "file:header_size" in asset_info.extra_fields:
            h = asset_info.extra_fields["file:header_size"]
            info["env"] = {"GDAL_INGESTED_BYTES_AT_OPEN": h}

        if requester_pays := extras["alternate"]["s3"].get("storage:requester_pays"):
            if requester_pays:
                info["env"] = {"AWS_REQUEST_PAYER": "requester"}

        if bands := extras.get("raster:bands"):
            stats = [
                (b["statistics"]["minimum"], b["statistics"]["maximum"])
                for b in bands
                if {"minimum", "maximum"}.issubset(b.get("statistics", {}))
            ]
            if len(stats) == len(bands):
                info["dataset_statistics"] = stats

        return info
```

moradology: My memory on this was fuzzy but I thought I recalled strange inheritance behavior with attr properties. I'll clean up the duplication

raster_api/runtime/src/alternate_reader.py
```diff
@@ -0,0 +1,55 @@
...
+        if asset not in self.assets:
+            raise InvalidAssetName(f"{asset} is not valid")
+
+        asset_info = self.input.assets[asset]
+        extras = asset_info.extra_fields
+
+        if ("alternate" not in extras) or ("s3" not in extras["alternate"]):
+            raise MissingAssets("No alternate asset found")
+
+        info = AssetInfo(url=extras["alternate"]["s3"]["href"], metadata=extras)
```
vincentsarago, line 34: in rio-tiler we use pystac to make sure the HREF is not relative https://github.com/cogeotiff/rio-tiler/blob/main/rio_tiler/io/stac.py#L298 but this being `custom` I think using direct link is fine 👍

vincentsarago (approved):

PR #284, opened 2024-01-30
test: weighted pixel area statistics

moradology:
This PR aims to provide tests for appropriate behavior of pixel-area-weighted statistics.

The basic idea behind this test is spelled out [here](https://gist.github.com/moradology/531a4b044350bebf5c0fe8877cb74f3a)


moradology:
It is looking as though I'm going to have to put in a bit of effort to get these integration tests to easily run locally. At the moment, the `docker compose` on our README is pretty inadequate for safely, conveniently verifying CI behavior

j08lue:
> At the moment, the docker compose on our README is pretty inadequate for safely, conveniently verifying CI behavior

Ah, please ticket that, @moradology.

moradology:
Issue lodged at #285

j08lue (changes requested):
This is a relevant test to have in our test suite, I think, so we can sustain trust in the accuracy of our zonal statistics method.

Can we generate the test file on the fly, rather than loading it from cloud storage? Then the test code would include information about how the data is created together with the expected result, which would make it more helpful and avoid relying on some storage that might get abandoned or so.

PR #303, opened 2024-02-15
fix: change search base condition causing key error

smohiudd:


stac_api/runtime/src/core.py
```diff
@@ -138,24 +138,27 @@ async def _search_base(
         result = await _super._search_base(search_request, **kwargs)
 
         if len(result["features"]) > 0:
```
smohiudd, line 140: Do we even need the if conditional if we're using `try`

anayeaye: I guess not but we could clean that up later

anayeaye (approved):

PR #305, opened 2024-02-15
fix(stac): hotfix stac-api change from main

anayeaye:
## What
Add this downstream fix https://github.com/NASA-IMPACT/veda-backend/pull/303 to develop


anayeaye:
this is not the right way to pull the hotfix changes in main

Issue #308, opened 2024-02-15
Update definition for IAM role for STAC Ingestor to be unique


botanical:
Currently, [the role is defined](https://github.com/NASA-IMPACT/veda-backend/blob/develop/ingest_api/infrastructure/construct.py#L130 ) to be `"delta-backend-staging-stac-ingestion-api-{stage}"` where `stage` is being set by https://github.com/NASA-IMPACT/veda-backend/blob/b5d5e3b6892e194abab88b77d607edc24df12fab/ingest_api/infrastructure/config.py#L13-L22 which will return `runner` as a the default. 

We need to update the construct to generate a unique construct name so that a unique logical ID is produced by CDK.

AC

- [ ] The code is updated so that a unique IAM role is created across staging environments
- [ ] Local testing is done in development

PR #309, opened 2024-02-16
fix: update iam role name for stac ingestor to be unique across stages

botanical:
### Issue
- https://github.com/NASA-IMPACT/veda-backend/issues/308

### Changes Made
- updated the role name to use `stack_name` since the original `stage` variable would default to `runner` across different stages causing deployments to `dev` to fail

### Testing Done
- Locally deployed to dev successfully and confirmed that the previous resource `delta-backend-staging-stac-ingestion-api-runner` still exists and that `stac-ingestion-api-veda-backend-uah-dev-role` was created and the `dev` deployment succeeded

### DIFF
--- a/ingest_api/infrastructure/construct.py
+++ b/ingest_api/infrastructure/construct.py
@@ -120,6 +120,7 @@ class ApiConstruct(Construct):
         db_subnet_public: bool,
         code_dir: str = "./",
     ) -> apigateway.LambdaRestApi:
+        stack_name = Stack.of(self).stack_name
         handler_role = iam.Role(
             self,
             "execution-role",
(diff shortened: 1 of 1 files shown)


botanical:
Pending a successful dev deployment when this gets merged, I will open a new PR that points to main to get this in staging as well.

anayeaye (approved):
I think we should add a tech debt task to our backlog to sort out how that config.stage is used and remove it if we can but this is great to see the CDK unblocked!

anayeaye:
> Pending a successful dev deployment when this gets merged, I will open a new PR that points to main to get this in staging as well.

Let's let that develop>main PR rest over the weekend, though. As in: let's not merge into main because it'd be deploying to staging (production) on a Friday.

botanical:
> I think we should add a tech debt task to our backlog to sort out how that config.stage is used and remove it if we can but this is great to see the CDK unblocked!

Good idea, @anayeaye !💡 I created the issue here https://github.com/NASA-IMPACT/veda-backend/issues/310 and added it to the backlog

Issue #291, opened 2024-02-05
Deploy dev veda backend with openveda.cloud


smohiudd:

botanical:
I created a certificate for the new domain `openveda.cloud` and tested that I configured correctly in dev with an updated deployment but I noticed the base url webpage throws an error. I discovered that the same error shows on https://dev.delta-backend.com/ so I'm working on debugging it.

<img width="1730" alt="Screenshot 2024-02-09 at 5 08 47 PM" src="https://github.com/NASA-IMPACT/veda-backend/assets/12633533/35c02162-b41b-4ee5-90f9-8df551b1af13">

botanical:
For anyone following along the issue, the following links are available:
https://dev.openveda.cloud/
https://dev.openveda.cloud/api/raster/docs
https://dev.openveda.cloud/api/stac/docs
https://dev.openveda.cloud/api/ingest/docs

I'm currently working on migrating to Stadia maps for https://dev.openveda.cloud/api/stac/index.html to be functional

botanical:
> For anyone following along the issue, the following links are available: https://dev.openveda.cloud/ https://dev.openveda.cloud/api/raster/docs https://dev.openveda.cloud/api/stac/docs https://dev.openveda.cloud/api/ingest/docs
> 
> I also will update the dev secret to contain the new values for the following variables, pending some confirmation that that's okay to do:
> 
> * VEDA_DOMAIN_HOSTED_ZONE_ID
> * VEDA_DOMAIN_HOSTED_ZONE_NAME
> * VEDA_CERT_ARN
> * VEDA_CUSTOM_HOST
> 
> I'm currently working on migrating to Stadia maps for https://dev.openveda.cloud/api/stac/index.html to be functional

I updated my comment and reverted the changes to the dev secret until I get confirmation from the team that it's okay to update it 😅

botanical:
The PR attached to this issue has been merged 🎉 Thanks for the review, @smohiudd!

Once we can successfully get dev deployed again, I will udpate the Stadia Maps account with the new domain (openveda.cloud)

botanical:
Writing here to note that I updated the Stadia Map account to use the new domain (openveda.cloud).

I also opened a PR to address an issue I found in the [STAC viewer](https://dev.openveda.cloud/api/stac/index.html) https://github.com/NASA-IMPACT/veda-backend/pull/311.

PR #311, opened 2024-02-17
fix: update stac viewer endpoint value

botanical:
### Issue
https://github.com/NASA-IMPACT/veda-backend/issues/291

### What?

- Updated the endpoint value for the STAC viewer in order to properly render the collections on the webpage. 
- Added a PR template so that all new PRs will have a template for reference

### Why?

- The endpoint is https://dev.openveda.cloud/collections which is incorrect which causes the collections filter to fail in the STAC viewer. 

https://github.com/NASA-IMPACT/veda-backend/assets/12633533/b272fb4c-23a9-41c2-8ce8-925c6760231a

### Testing?

- I didn't locally deploy or anything but here is [the working collections link](https://dev.openveda.cloud/api/stac/collections)

### Notes
- I think we could technically also update https://github.com/NASA-IMPACT/veda-backend/blob/develop/stac_api/runtime/src/app.py#L79 to be:
```{"request": request, "endpoint": str(request.url).replace("/index.html", "/api/stac")},``` if that's preferred

### DIFF
--- a/stac_api/runtime/src/templates/stac-viewer.html
+++ b/stac_api/runtime/src/templates/stac-viewer.html
@@ -662,7 +662,7 @@ const switchPane = (event) => {
 
 map.on('load', () => {
   // get collections
-  fetch('{{ endpoint }}/collections')
+  fetch('{{ endpoint }}/api/stac/collections')
     .then(res => {
       if (res.ok) return res.json()
       throw new Error('Network response was not ok.')
--- a/stac_api/runtime/src/templates/stac-viewer.html
+++ b/stac_api/runtime/src/templates/stac-viewer.html
@@ -662,7 +662,7 @@ const switchPane = (event) => {
 
 map.on('load', () => {
   // get collections
-  fetch('{{ endpoint }}/api/stac/collections')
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: stac_api/runtime/src/templates/stac-viewer.html [hunk 1]
```diff
@@ -662,7 +662,7 @@
 
 map.on('load', () => {
   // get collections
-  fetch('{{ endpoint }}/collections')
+  fetch('{{ endpoint }}/api/stac/collections')
```
anayeaye, line 665: this is configurable so we should try to pass the config.py `root_path` here if possible (there may be an example of doing this with the template for the cog viewer in the raster api). If not we should capture the need in our back log so we are not surprised in the future if veda-backend deployed with a different root path

botanical: https://github.com/NASA-IMPACT/veda-backend/pull/311/commits/bdd898cc4f55e984a6971184f934b60c7edf9374

anayeaye (approved):

PR #317, opened 2024-02-21
fix: support automatic setting of search_path

moradology:
This is a workaround for the error being encountered by titiler-pgstac's use of command line arguments. Application can be set via query paramaters and search_path is set by automatically executing `SET search_path` upon acquisition of connection

### DIFF
--- a/raster_api/runtime/src/app.py
+++ b/raster_api/runtime/src/app.py
@@ -6,6 +6,7 @@ from aws_lambda_powertools.metrics import MetricUnit
 from src.algorithms import PostProcessParams
 from src.alternate_reader import PgSTACReaderAlt
 from src.config import ApiSettings
+from src.db import connect_to_rds_proxy
 from src.dependencies import ColorMapParams, ItemPathParams
 from src.extensions import stacViewerExtension
 from src.monitoring import LoggerRouteHandler, logger, metrics, tracer
@@ -39,11 +40,13 @@ else:
     optional_headers = []
 
 
+await connect_to_db()
+
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     """FastAPI Lifespan."""
     # Create Connection Pool
-    await connect_to_db(app, settings=settings.load_postgres_settings())
+    await connect_to_rds_proxy(app, settings=settings.load_postgres_settings())
     yield
     # Close the Connection Pool
     await close_db_connection(app)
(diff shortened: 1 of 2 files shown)


vincentsarago (approved):

anayeaye (commented):
Quick note: Please do not merge without coordinating in the data services channel. We currently have several test operations competing for the dev infrastructure so we need to coordinate.

vincentsarago:
FYI, another solution (once https://github.com/stac-utils/titiler-pgstac/pull/155 get's published) would have been to use a custom mosaic backend 

```python

from titiler.pgstac import mosaic

@attr.s
class PGSTACBackend(mosaic.PGSTACBackend)
    @cached(  # type: ignore
        TTLCache(maxsize=cache_config.maxsize, ttl=cache_config.ttl),
        key=lambda self, geom, **kwargs: hashkey(self.input, str(geom), **kwargs),
    )
    @retry(
        tries=retry_config.retry,
        delay=retry_config.delay,
        exceptions=(
            pgErrors.OperationalError,
            pgErrors.InterfaceError,
        ),
    )
    def get_assets(
        self,
        geom: Geometry,
        fields: Optional[Dict[str, Any]] = None,
        scan_limit: Optional[int] = None,
        items_limit: Optional[int] = None,
        time_limit: Optional[int] = None,
        exitwhenfull: Optional[bool] = None,
        skipcovered: Optional[bool] = None,
    ) -> List[Dict]:
        """Find assets."""
        fields = fields or {
            "include": ["assets", "id", "bbox", "collection"],
        }

        scan_limit = scan_limit or 10000
        items_limit = items_limit or 100
        time_limit = time_limit or 5
        exitwhenfull = True if exitwhenfull is None else exitwhenfull
        skipcovered = True if skipcovered is None else skipcovered

        with self.pool.connection() as conn:
            with conn.cursor() as cursor:
                try:
                    cursor.execute("SET search_path TO pgstac, public")
                    cursor.execute(
                        "SELECT * FROM geojsonsearch(%s, %s, %s, %s, %s, %s, %s, %s);",
                        (
                            geom.model_dump_json(exclude_none=True),
                            self.input,
                            json.dumps(fields),
                            scan_limit,
                            items_limit,
                            f"{time_limit} seconds",
                            exitwhenfull,
                            skipcovered,
                        ),
                    )
                    resp = cursor.fetchone()[0]

                except pgErrors.RaiseException as e:
                    # Catch Invalid SearchId and raise specific Error
                    if f"Search with Query Hash {self.input} Not Found" in str(e):
                        raise MosaicNotFoundError(
                            f"SearchId `{self.input}` not found"
                        ) from e
                    else:
                        raise e

        return resp.get("features", [])
```

moradology:
Looks to me like bumping titiler is the path forward here; closing this for now

PR #320, opened 2024-02-23
fix(routes): make route behaviors additive

ividito:
https://github.com/NASA-IMPACT/veda-data-airflow/pull/108 adds a cloudfront behavior that we don't want to maintain here. This is an attempt to avoid clobbering that deployment whenever our backend gets deployed.

### DIFF
--- a/routes/infrastructure/construct.py
+++ b/routes/infrastructure/construct.py
@@ -58,27 +58,30 @@ class CloudfrontDistributionConstruct(Construct):
                 domain_names=[f"{stage}.{veda_route_settings.domain_hosted_zone_name}"]
                 if veda_route_settings.domain_hosted_zone_name
                 else None,
-                additional_behaviors={
-                    "/api/stac*": cf.BehaviorOptions(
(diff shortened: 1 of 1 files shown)


routes/infrastructure/construct.py
```diff
@@ -58,27 +58,30 @@ def __init__(
...
-                        origin=origins.HttpOrigin(
-                            f"{stac_api_id}.execute-api.{region}.amazonaws.com",
-                            origin_id="stac-api",
-                        ),
-                        cache_policy=cf.CachePolicy.CACHING_DISABLED,
-                        allowed_methods=cf.AllowedMethods.ALLOW_ALL,
-                        origin_request_policy=cf.OriginRequestPolicy.ALL_VIEWER_EXCEPT_HOST_HEADER,
+            )
+
+            if veda_route_settings.cloudfront:
```
anayeaye, line 63: This block is already inside of a if cloudfront condition (line 35)

anayeaye (approved):

PR #323, opened 2024-02-26
feat: configurable subnet ids

slesaad:
### What?

- Allows users to specify the subnet IDs to be used for the database

### Why?

- Since we don't always get to create our own VPC, the subnets associated with it are not always one or the other type. We've had issues in the past with the private subnet type not being available in the VPC (`PRIVATE_ISOLATED`, `PRIVATE_EGRESS`, etc. This allows users to specify the subnet ids so we don't rely on a specific type.

### Testing?

- Deployed and manually checked that the database and the ingest lambdas were associated to the subnets provided

### DIFF
--- a/database/infrastructure/construct.py
+++ b/database/infrastructure/construct.py
@@ -156,10 +156,14 @@ class RdsConstruct(Construct):
         )
 
         # Configure accessibility
-        if (subnet_ids):
+        if subnet_ids:
             self.vpc_subnets = aws_ec2.SubnetSelection(
-                subnets=[aws_ec2.Subnet.from_subnet_attributes(self, f"Subnet{i}", subnet_id=subnet_id)
-                         for i, subnet_id in enumerate(subnet_ids)]
+                subnets=[
+                    aws_ec2.Subnet.from_subnet_attributes(
+                        self, f"Subnet{i}", subnet_id=subnet_id
+                    )
+                    for i, subnet_id in enumerate(subnet_ids)
+                ]
             )
         else:
             subnet_type = (
--- a/database/infrastructure/construct.py
+++ b/database/infrastructure/construct.py
@@ -1,7 +1,7 @@
 """CDK Construct for veda-backend RDS instance."""
 import json
 import os
-from typing import Union
+from typing import List, Optional, Union
 
 from aws_cdk import (
     CfnOutput,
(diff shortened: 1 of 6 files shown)


database/infrastructure/construct.py
```diff
@@ -162,19 +155,38 @@ def __init__(
...
+                ]
+            )
+        else:
+            subnet_type = (
+                aws_ec2.SubnetType.PRIVATE_WITH_EGRESS
+                if not veda_db_settings.publicly_accessible
+                else aws_ec2.SubnetType.PUBLIC
+            )
+            self.vpc_subnets = aws_ec2.SubnetSelection(subnet_type=subnet_type)
+
```
slesaad, line 175: maintain backward compatibility via the else block

ingest_api/infrastructure/config.py
```diff
@@ -36,18 +34,9 @@ class IngestorConfig(BaseSettings):
...
-        description="Name of secret containing pgSTAC DB connection information"
-    )
-    stac_db_vpc_id: str = Field(description="ID of VPC running pgSTAC DB")
     stac_db_security_group_id: str = Field(
         description="ID of Security Group used by pgSTAC DB"
     )
-    stac_db_public_subnet: bool = Field(
-        description="Boolean indicating whether or not pgSTAC DB is in a public subnet",
-        default=True,
-    )
```
slesaad, line 50: these are not being used anymore, but directly passed to the Ingestor Constructs

ingest_api/runtime/src/config.py
```diff
@@ -4,7 +4,6 @@
 from pydantic_ssm_settings import AwsSsmSourceConfig
 
 AwsArn = constr(regex=r"^arn:aws:iam::\d{12}:role/.+")
-AwsStepArn = constr(regex=r"^arn:aws:states:.+:\d{12}:stateMachine:.+")
```
slesaad, line 7: not used

ingest_api/infrastructure/construct.py
```diff
@@ -296,7 +293,7 @@ def build_ingestor(
         db_secret: secretsmanager.ISecret,
         db_vpc: ec2.IVpc,
         db_security_group: ec2.ISecurityGroup,
-        db_subnet_public: bool,
```
slesaad, line 299: this is no longer required since we already account for `db_publicly_accessible` from the database when selecting the subnets

PR #324, opened 2024-02-26
fix(ci): updates for veda deploy

slesaad:
### Issue

https://github.com/NASA-IMPACT/veda-deploy/issues/25

### What?

- Updates needed to deploy using `veda-deploy`
- Update version of GitHub actions

### Why?

- for `veda-deploy` to be able to pass env vars from one stack to another

### Testing?

- Ran deployment using `veda-deploy` until successful

### DIFF
--- a/.github/actions/cdk-deploy/action.yml
+++ b/.github/actions/cdk-deploy/action.yml
@@ -4,8 +4,9 @@ inputs:
   env_aws_secret_name:
     required: true
     type: string
-  auth_env_aws_secret_name:
-    required: true
+  script_path:
+    type: string
+  auth_stack_name:
     type: string
   dir:
     required: false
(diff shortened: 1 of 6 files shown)


anayeaye:
This is different in the [pre-deploy cdk diff action](https://github.com/NASA-IMPACT/veda-backend/actions/runs/8056366338/job/22006048164#step:11:1522): `Could not create a change set, will base the diff on template differences (run again with -v to see the reason)`. I can re-run the action to see if that is a transient message but if not I guess we should update pr.yml to be more verbose `cdk diff -v --outputs-file ${HOME}/cdk-outputs.json`. 

I'm otherwise ready to approve, just want to see what is up with the cdk diff set.

anayeaye:
I made the cdk more verbose and now we can see that it tries to create the change set a few times before succeeding. So apparently it was a transient error but I am glad to see the change set now.

Issue #297, opened 2024-02-07
Dependabot Alerts


anayeaye:
## What
https://github.com/NASA-IMPACT/veda-backend/security/dependabot/1

https://github.com/NASA-IMPACT/veda-backend/security/dependabot/2

## Note
We've pinned starlette in the raster-api on an earlier version to avoid this `[ERROR] AssertionError: either 'directory' or 'env' arguments must be passed`

AC

- [x] deployed to dev
- [x] veda-config check

smohiudd:
@vincentsarago what's the recommended version of starlette we should be using for titiler? We've currently pinned `starlette<0.28`

vincentsarago:
@smohiudd tbh I always deferred the starlette pining to fastAPI so it depends on which titiler version you use

PR #335, opened 2024-03-13
fix: api dependencies upgrade

anayeaye:
# fixed
- python-multipart upgraded for security
- starlette version un-pinned (let fastapi manage this)

# ci
- all github actions upgraded to resolve node outdated node warnings
- conventional pr title linting action replaced with https://github.com/amannn/action-semantic-pull-request which can run on node > 16
- release tag generated before deployment in both develop and main workflows (makes it easier to re-run last deployment with configuration changes and should add the git release tag to the tags on aws resources)

## Notes
Two dependabot alerts point back to the same python-multipart vulnerability, the second suggest upgrading starlette to version [0.36.2](https://github.com/encode/starlette/releases/tag/0.36.2) (the first release that pins python-multipart version at all). This PR instead requires `python-multipart>=0.0.7` which will be compatible with older versions of starlette needed by our other dependencies.

## issue
#297

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -71,8 +71,27 @@ jobs:
       - name: Stop services
         run: docker compose stop
 
-  deploy-dev:
+  pre-release:
     needs: [test-dev]
+    runs-on: ubuntu-latest
+    concurrency: release
+    permissions:
+      id-token: write
+      contents: write
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Python Semantic Release
+      uses: python-semantic-release/python-semantic-release@master
+      with:
+        changelog: "false"
+        vcs-release: "true"
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+
+  deploy-dev:
+    needs: [pre-release]
 
     runs-on: ubuntu-latest
     steps:
@@ -120,22 +139,3 @@ jobs:
         run: |
           echo $STAGE
           cdk deploy --require-approval never --outputs-file ${HOME}/cdk-outputs.json
-
-  pre-release:
-    needs: [deploy-dev]
-    runs-on: ubuntu-latest
-    concurrency: release
-    permissions:
-      id-token: write
-      contents: write
-
-    steps:
-    - uses: actions/checkout@v4
-      with:
-        fetch-depth: 0
-    - name: Python Semantic Release
-      uses: python-semantic-release/python-semantic-release@master
-      with:
-        changelog: "false"
-        vcs-release: "true"
-        github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


raster_api/runtime/setup.py
```diff
@@ -14,7 +14,7 @@
     "starlette-cramjam>=0.3,<0.4",
     "aws_xray_sdk>=2.6.0,<3",
     "aws-lambda-powertools>=1.18.0",
-    "starlette<0.28",
+    "python-multipart==0.0.7",
```
botanical, line 17: Why this version? 😯

CONTRIBUTING.md
```diff
@@ -14,7 +14,7 @@ This project generally follows the [git-flow branching model](https://nvie.com/p
 
 - Pull requests should explain **what** is changed, **why**, and **how the change can be tested**.
   
-- Pull request *titles* must adhere to the [conventional commits specification]([#conventional-commits](https://www.conventionalcommits.org/en/v1.0.0/)). This is enforced using the  [conventional-pull-request-action](https://github.com/CondeNast/conventional-pull-request-action).  
+- Pull request *titles* must adhere to the [conventional commits specification]([#conventional-commits](https://www.conventionalcommits.org/en/v1.0.0/)). This is enforced using [action-semantic-pull-request](https://github.com/amannn/action-semantic-pull-request).
```
botanical, line 17: 👍

.github/workflows/develop.yml
```diff
@@ -71,22 +71,41 @@ jobs:
...
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: '3.9'
 
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: 17
+          node-version: 20
```
botanical, line 108: ⭐

raster_api/runtime/setup.py
botanical: Also, why remove starlette? 🤔

botanical: > Also, why remove starlette? 🤔

Ah, is it because of the dependabot alert?

botanical (approved):
develop and main yaml file changes are consistent and I see that the pr.yml file was updated to use action-semantic-pull-request. Changes look good!

raster_api/runtime/setup.py
vincentsarago: Starlette version is managed by FastAPI module directly (so by titiler version), so having this here is redondant and possibly dangerous.

anayeaye: @vincentsarago we're pinning the min python-multipart version and removing any pin on starlette here and letting other packages manage the version. Do you think it is also a problem to pin python-multipart?

anayeaye: Oh I just re-read the full thread. I think we are good and in agreement 😅

smohiudd (approved):

PR #338, opened 2024-03-15
fix: stac browser catalog url with cloudfront

anayeaye:
## What
Force a trailing slash on the stac catalog url when building a stac-browser for a stac-api behind a cloudfront.

## How tested
I temporarily updated the deploy action to push to the dev stack on push to this fix branch and confirmed that the browser at dev.openveda.cloud properly finds the stac api at dev.openveda.cloud/api/stac/.

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
     - develop
+    - fix/stac-browser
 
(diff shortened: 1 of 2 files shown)


anayeaye:
Screen Shot 2024-03-15 at 4 00 32 PM

smohiudd (approved):

PR #340, opened 2024-03-18
ci: update predeploy actions

anayeaye:
## What
updates more github action versions


botanical (approved):
good catch, @anayeaye

Issue #289, opened 2024-02-01
Implement OpenTelemetry in Ingest API


smohiudd:

smohiudd:
https://github.com/NASA-IMPACT/veda-backend/pull/342

PR #342, opened 2024-03-22
feat(otel): add otel to ingest API

ividito:
### Issue

#289 

### What?

- Adds basic monitoring functionality to ingest API

### Testing?

- Deployed to temporary `otel` stack and viewed generated metrics and logs using POC Grafana instance

### DIFF
--- a/ingest_api/runtime/src/main.py
+++ b/ingest_api/runtime/src/main.py
@@ -7,13 +7,16 @@ import src.config as config
 import src.dependencies as dependencies
 import src.schemas as schemas
 import src.services as services
+from aws_lambda_powertools.metrics import MetricUnit
 from src.collection_publisher import CollectionPublisher, ItemPublisher
 from src.doc import DESCRIPTION
+from src.monitoring import LoggerRouteHandler, logger, metrics, tracer
 
-from fastapi import Depends, FastAPI, HTTPException
+from fastapi import APIRouter, Depends, FastAPI, HTTPException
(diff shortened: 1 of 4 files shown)


ingest_api/runtime/src/main.py
```diff
@@ -222,7 +227,42 @@ def who_am_i(claims=Depends(auth.decode_token)):
...
 async def validation_exception_handler(request, exc):
     return JSONResponse(str(exc), status_code=422)
+
+
+@app.exception_handler(Exception)
+async def general_exception_handler(request, err):
+    """Handle exceptions that aren't caught elsewhere"""
+    metrics.add_metric(name="UnhandledExceptions", unit=MetricUnit.Count, value=1)
+    logger.exception("Unhandled exception")
+    return JSONResponse(status_code=500, content={"detail": "Internal Server Error"})
```
anayeaye, line 268: Can we log the err here? `f"Unhandled exception {err=}"`?

anayeaye (approved):
Approve but I am curious about the unhandled/unlogged exception err.

anayeaye:
@smohiudd @ividito I wonder if we should push this through before the auth PR #356 (we also may have to update the auth branch to resolve any diffs).

ingest_api/runtime/src/main.py
smohiudd: I think it makes sense to log the error

PR #344, opened 2024-03-26
fix(db)!: update collection summaries on load items

anayeaye:
## What
- Improve veda loader method that updates the summaries and extents of all items in a collection with a new `collection_temporal_extent_max` function that includes the maximum item `end_datetime` which is significantly different from an item's nominal `datetime` in collections of items that represent an entire year or range or dates.
- Removes ingest api large number test that is no longer needed because we no longer convert json values for the DDB schema.
- Pins fastapi at for ingest api tests (tech debt). Note that python-multipart is pinned at the new and secure version so it is still OK for us to use older versions of fastapi and starlette for now.
- Sets moto version to fix tests.

> **Note**
This requires incrementing `VEDA_DB_SCHEMA_VERSION` to cause the dashboard schema update to run so I am making it a breaking change.

## How tested
I removed the latest date from the geoglam collection's temporal extent _and_ datetime summaries then ingested a single item to /ingestions which triggered both the summaries and extent functions to be executed. Verified in dev.openveda.cloud/api/stac/collections/geoglam.

### DIFF
--- a/ingest_api/runtime/tests/test_registration.py
+++ b/ingest_api/runtime/tests/test_registration.py
@@ -9,7 +9,6 @@ import pytest
 
 from fastapi.encoders import jsonable_encoder
 
-# from src.schemas import Ingestion
 
 if TYPE_CHECKING:
     from src import schemas, services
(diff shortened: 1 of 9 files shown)


anayeaye:
Ingest api tests are now failing with the same moto version issue that @stephenkilbourn has been fighting I think. I will dig into that next, too.

ingest_api/runtime/tests/test_registration.py
```diff
@@ -1,11 +1,13 @@
 import base64
 import json
 from datetime import timedelta
-from math import isclose
+
+# from math import isclose
```
stephenkilbourn, line 5: Should this import line just be deleted?

stephenkilbourn (approved):

Issue #347, opened 2024-04-05
Implement ingest-api auth login from swagger


anayeaye:
## What
Currently administrators must manually post username and password to a token endpoint and copy paste the token from the response for ingest operations. Update this auth flow to follow the more standard redirect to auth provider for a secure username and password form entry and redirect to the swagger docs.

## AC
- [ ] token auth urls are configurable by veda environment
- [ ] admins in the cognito user pool associated with backend stack can login via swagger docs
- [ ] whoami response demonstrates the token is valid
- [ ] token endpoint removed if no longer functional

anayeaye:
UPDATE

**Working** 
The `feature/ingest_api/docs-auth-flow` branch now 
1. collects the cognito domain from the environment (this is the pattern expected for veda-deploy which reads the cognito programmatic client secret and exports to the runner environment)
2. The ingest api now infers the authorization and token urls from that cognito domain
3. If the user pool is manually updated with the ingest API url, then when a user clicks the authorize button in the swagger docs they are redirected temporarily to the cognito hosted UI to provide username and password and the directed back to the docs.

**Needs work**
I put some time into programmatically updating the user pool client's hosted UI to add the ingest-api docs callback url using a custom resource + AwsSdkCall. I have backed these changes out to reconsider the approach.

Issue [comment](https://github.com/aws/aws-cdk/issues/3037#issuecomment-601427972) that inspired the spike
[AWS Custom Resource**s** SdkCall](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.custom_resources/AwsSdkCall.html)
[Update user pool client docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp/client/update_user_pool_client.html#)

This snippet adds the desired callback url but stomps the rest of the hosted ui configuration including existing callback urls
```
from aws_cdk import custom_resources as cr
# Append api to cognito client allowed callback urls with an AwsSdkCall
        stack_name = Stack.of(self).stack_name
        oauth_redirect_url = f"{self.api.url.rstrip('/')}/docs/oauth2-redirect"
        callback_urls = [oauth_redirect_url]
        cr.AwsCustomResource(
            self,
            id="UpdateClientCallbackUrls",
            function_name=f"{stack_name}-UpdateClientCallbackUrls",
            policy=cr.AwsCustomResourcePolicy.from_sdk_calls(
                resources=cr.AwsCustomResourcePolicy.ANY_RESOURCE
[...]

PR #365, opened 2024-04-29
fix: update get_username function 

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-data-airflow/issues/134
https://github.com/NASA-IMPACT/veda-backend/issues/347

### What?

- Update get_username to fall back on `sub` if `username` doesn't exist in token
- Added some more logging statements to make debugging easier in the future

### Why?

- This fix is to enable a successful workflows API run since the workflow API passes a token to backend API and it's currently erroring
- Also, `sub` is a more definitive identifier because it represents a unique identifier compared to `username` 

### Testing?

- Relevant testing details

### DIFF
--- a/ingest_api/runtime/src/auth.py
+++ b/ingest_api/runtime/src/auth.py
@@ -58,6 +58,8 @@ def validated_token(
 def get_username(token: Annotated[Dict[Any, Any], Depends(validated_token)]) -> str:
     logger.info(f"\nToken {token}")
     result = token["username"] if "username" in token else token.get("sub", None)
+    if result is None:
+        raise KeyError(f"Neither 'username' nor 'sub' found in the token: {token}.")
     return result
 
 
--- a/ingest_api/runtime/src/auth.py
+++ b/ingest_api/runtime/src/auth.py
@@ -33,7 +33,7 @@ def validated_token(
             jwks_client.get_signing_key_from_jwt(token_str).key,
             algorithms=["RS256"],
         )
-        logger.info(f"\Decoded token {token}")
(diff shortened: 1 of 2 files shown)


smohiudd (approved):

ingest_api/runtime/src/auth.py
```diff
@@ -53,8 +55,10 @@ def validated_token(
     return token
 
 
-def get_username(token: Annotated[Dict[Any, Any], Depends(validated_token)]):
-    return token["username"]
+def get_username(token: Annotated[Dict[Any, Any], Depends(validated_token)]) -> str:
+    logger.info(f"\nToken {token}")
+    result = token["username"] if "username" in token else token.get("sub", None)
```
anayeaye, line 60: I'm mixed about having a default when we might want to fail but because this depends on validation I think it is OK? We can decide after it works or not

anayeaye (approved):

ingest_api/runtime/src/auth.py
botanical: https://github.com/NASA-IMPACT/veda-backend/pull/365/commits/dbb62fbc15ce309ef5d12abe8bbee600c66a48ac Perhaps we can raise the error instead?

botanical: [f640fdf](https://github.com/NASA-IMPACT/veda-backend/pull/365/commits/f640fdf537e8d56cfd4e8321be0e25e97e029df1) change reverted since we don't want to be blocked and could forget about this change later on

PR #369, opened 2024-05-06
fix(ingest): default empty string for root path in ingest api construct

anayeaye:
### What?
- Allow ingest-api to be deployed without a custom root path by setting the default config for fastapi root path parameter to None.

### Why?
- When the ingest API is deployed to a stack without a cloud front and custom host it should not apply an additional root path.

### How tested
Deployed to a test branch using a custom subdomain instead of a cloudfront with custom root path.


botanical (approved):
Changes are consistent with what we have configured for raster and stac so this looks good! I know this PR is going to be tested ad-hoc well so thanks 🙇

PR #376, opened 2024-05-15
feat!: configure cloudfront with origin access control (proposal)

botanical:
### Issue

No relevant github issue but in MCP, setting a default root object is required which this work helps to address. 

### What?

- enable logging (this was added in [another merged PR](https://github.com/NASA-IMPACT/veda-backend/pull/375))
- configure cloudfront to use origin access control and delete the origin access identity that gets created
- set default root object to `index.html`
- use flag to enable/disable feature`VEDA_CLOUDFRONT_OAC`
### Why?

- currently, our MCP deployments need to be manually updated in order for the cloudfront distribution to work properly with S3 buckets that block public access

### Testing?

- deployed these changes to UAH dev and MCP test

### Other
- I opted to not add policy configuration in this PR since there is no way to tell via CDK if a policy on a bucket already exists. In order for this to work, the S3 browser bucket must allow cloudfront to `GetObject`
```
{
    "Version": "2008-10-17",
    "Id": "PolicyForCloudFrontPrivateContent",
    "Statement": [
        {
            "Sid": "AllowCloudFrontServicePrincipal",
            "Effect": "Allow",
            "Principal": {
                "Service": "cloudfront.amazonaws.com"
            },
            "Action": "s3:GetObject",
            "Resource": "arn:aws:s3:::BUCKET/*",
            "Condition": {
                "StringEquals": {
                    "AWS:SourceArn": "arn:aws:cloudfront::ACCOUNT:distribution/DISTRIBUTION_ID"
                }
            }
        }
    ]
```

### Misc
In order to properly configure this for our UAH stacks, we will need to update the buckets to block public access, not use static website hosting, and to include the bucket policy like above. These changes are required because our current UAH buckets do use static website hosting and therefore don't have the above policy and allow read access.


anayeaye (approved):
Let's make this a breaking `feat!` upgrade to alert maintainers of current implementations to either disable the `VEDA_CLOUDFRONT_OAC` or modify existing browser bucket configuration.

slesaad (approved):

PR #388, opened 2024-06-03
fix(ingest): requester pays config in validation

anayeaye:
### Issue

https://github.com/NASA-IMPACT/veda-data/issues/134

### What?

- Add optional requester pays configuration to ingest API to validate accessibility of assets in buckets that require requester pays _if_ the titiler been configured to use requester pays

### Why?

- Ingest API verifies that all hrefs in new items are accessible to the titiler but is failing in cases where the bucket requires requester pays configuration

### Testing?

I deployed this manually to dev and confirmed that the change resolves the unable to access objects errors when configured to use requester pays.

### DIFF
--- a/ingest_api/runtime/src/auth.py
+++ b/ingest_api/runtime/src/auth.py
@@ -26,7 +26,6 @@ def validated_token(
     required_scopes: security.SecurityScopes,
 ) -> Dict:
     # Parse & validate token
-    logger.info(f"\nToken String {token_str}")
     try:
         token = jwt.decode(
             token_str,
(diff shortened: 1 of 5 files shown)


ingest_api/runtime/src/auth.py
```diff
@@ -26,7 +26,6 @@ def validated_token(
     required_scopes: security.SecurityScopes,
 ) -> Dict:
     # Parse & validate token
-    logger.info(f"\nToken String {token_str}")
```
botanical, line 29: 🙇

botanical (approved):

Issue #384, opened 2024-05-28
Testing improvement in veda-backend


smohiudd:
## What

The [VEDA testing plan](https://docs.google.com/document/d/1EmSWgbCycIuXZ3CujI5J3L1fB1mp2MUXl3do3R9zass/edit) outlines the current state of testing coverage for backend services. The actions from the plan include the following:

- Determine custom endpoints not currently in STAC FastAPI or Titiler.
- Add tests to validate custom endpoints.
- Consider adding tests to check each of the routes through the api gateway, e.g. call GET /mosaic/{searchid}/tiles/{z}/{x}/{y}/asset and verify status code. This would validate successful deployment and guard against issues with cloudfront 

## PI Objective

Objective 6: https://github.com/NASA-IMPACT/veda-architecture/issues/410

## Acceptance Criteria

- [ ] Tested included for custom endpoints and deployed to dev

smohiudd:
Not maintained endpoints:
STAC: collectionid 
RASTER: alternate assets

PR #397, opened 2024-07-17
chore: Improve raster integration tests

stephenkilbourn:
### Issue

#384 

### What?
This updates the existing raster integration tests to cover more of the endpoints listed in the docs. It also pulls down the Openapi Schemas from https://api.stacspec.org/v1.0.0 and validates the feature and collection results match those schemas.

### Testing?
this will run with the existing script: ./scripts/run-local-tests.sh


.github/workflows/tests/test_raster.py
```diff
@@ -3,6 +3,9 @@
 import httpx
 
 raster_endpoint = "http://0.0.0.0:8082"
+seeded_collection = "noaa-emergency-response"
+seeded_id = "20200307aC0853900w361030"
+z, x, y = 15, 8589, 12849
```
ividito, line 8: Thanks for implementing this - would this PR be a good time to define these constants as [fixtures](https://docs.pytest.org/en/stable/how-to/fixtures.html)? Later on, when we implement more automated integration tests, we could use these to test live environments by overriding the endpoint and items for the tests.

stephenkilbourn: Thanks for the guidance. I think the updates are there now

anayeaye:
nit: can we bring these item tests back next time we are working in raster API? I'm pretty sure the dashboard uses these `/collections/{collection_id}/items/{item_id}` endpoints and I know the notebooks rely on them

https://github.com/NASA-IMPACT/veda-backend/pull/397/files#diff-131e458a6870e9c6459ad7e9fbbaae846353b8ee197693d81683126c74744bf8L174-L195

Issue #372, opened 2024-05-08
Make ingest-api pypgstac version configurable


anayeaye:
## What
Pypgstac version should correspond to the configured `PGSTAC_VERSION` but is currently hard coded in the ingest-api requirements.txt. See the database/runtime/Dockerfile for how the environment variable can be used to install the same pypgstac version as the pgstac version used in postgres.

`RUN pip install requests "urllib3<2" psycopg[binary,pool] pypgstac==${PGSTAC_VERSION} -t /asset`

## AC
- [ ] the version of `pypgstac` used in the ingest-api corresponds to the pgstac version in the database.

Issue #12, opened 2022-01-19
Implement RDS Paramaters for PGSTAC database


leothomas:
Parameters to set: 
- `max_locks_per_transaction: 1024`
- `work_mem: 64000`
- `temp_buffers: 32000`

Note: these parameters need to be further researched and fine tuned in accordance with usage patterns

## Acceptance Criteria
- [ ] Updates made are documented for the team to review (or even a command that somebody could use to look this up)
- [ ] These parameters are in the CDK

PR #404, opened 2024-07-29
chore: Make pypgstac an ENV in dockerfile

sandrahoang686:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/372

### What?

Make pypgstac and ENV var in dockerfile

### Why?

- Description of why the changes were made

### Testing?

Deployed without any errors to http://monitor.openveda.cloud/


ingest_api/runtime/Dockerfile
```diff
@@ -1,9 +1,12 @@
 FROM public.ecr.aws/sam/build-python3.9:latest
 
+ENV VEDA_DB_PGSTAC_VERSION=0.7.4
```
anayeaye, line 3:
```suggestion
ARG PGSTAC_VERSION
RUN echo "Using PGSTAC Version ${PGSTAC_VERSION}"
```
I think this should be coming directly from the build environment so we can try removing this line entirely. If not we may need add pgstac version to the via the ingest_api/infrastructure.

I think it will work like this but it may need a little trial and error:

1. For the config can mostly follow this example from the [database construct](https://github.com/NASA-IMPACT/veda-backend/blob/develop/database/infrastructure/config.py#L23-L26). 
2. But we'll need to call it db_pgstac_version in ingest_api/infrastructure/config.py if we want to pick up the same environment variable in the ingest api scope (some configs are prefixed with their scope)
3. To get the config into the ingestor, we'll can add it as a build arg for the [ingestor construct](https://github.com/NASA-IMPACT/veda-backend/blob/develop/ingest_api/infrastructure/construct.py#L316) that uses pypgstac. There's a [build arg](https://github.com/NASA-IMPACT/veda-backend/blob/develop/database/infrastructure/construct.py#L55) example in the database construct.

ingest_api/runtime/requirements.txt
```diff
@@ -8,7 +8,7 @@ orjson>=3.6.8
 psycopg[binary,pool]>=3.0.15
 pydantic_ssm_settings>=0.2.0
 pydantic>=1.10.12
-pypgstac==0.7.4
+pyjwt>=2.8.0
```
anayeaye, line 11: this came from merging upstream develop changes. The same line and same first two characters confused the merge conflict resolver

anayeaye:
@sandrahoang686 I took the liberty of pulling recent changes from develop into your branch and making [this small change ](https://github.com/NASA-IMPACT/veda-backend/pull/404/commits/ea88a2c082c20de38a893d22c23ff2e5f8bd2f2a)to get the pgstac_version in the build api parameters. I can see that worked in the action because the docker build is now installing the correct version [`#12 [ 8/16] RUN pip install -r /tmp/ingestor/requirements.txt pypgstac==0.7.10 -t /asset --no-binary pydantic uvicorn`](https://github.com/NASA-IMPACT/veda-backend/actions/runs/10309464549/job/28539342322#step:11:1072). Now I see that we need to do the same for the ingestor construct (`build_ingestor`)...

ingest_api/runtime/requirements.txt
sandrahoang686: thanks for pointing this out! I got rid of `pyjwt`.. because looks like it was no longer in the `develop` branch and added `pypgstac` back

sandrahoang686:
@anayeaye I've passed in `pgstac_version` to ingest_api => `build_ingestor` and added it as a `build_arg` to `docker_build`. Am i missing anything? If I wanted to test this btw, how would I do so correctly?

ingest_api/runtime/requirements.txt
anayeaye: Yeah I guess we get it here now https://github.com/NASA-IMPACT/veda-backend/blob/develop/common/auth/setup.py#L6

anayeaye: But we still do need `- pypgstac==0.7.4`

anayeaye (approved):
I see the correct pypgstac installed in the github action log as well as Code updates for both the ingest api handler and the ingestor construct.

anayeaye:
> @anayeaye I've passed in `pgstac_version` to ingest_api => `build_ingestor` and added it as a `build_arg` to `docker_build`. Am i missing anything? If I wanted to test this btw, how would I do so correctly?

@sandrahoang686 you could do a local cdk synth and look at the outputs but I think the easiest way to verify this kind of change is to read the pre-deploy job associated with this PR. I searched to be sure that the correct pypgstac was installed in both ingest lambda Dockerfile builds.

smohiudd (approved):

Issue #402, opened 2024-07-22
Necessary values are propagated to veda-routes repo for CloudFront distribution origin management


botanical:
## Context

- Epic Link: https://github.com/NASA-IMPACT/veda-architecture/issues/469
- ADR: https://github.com/NASA-IMPACT/veda-architecture/pull/464

Propagate the necessary values for CloudFront distribution origin management 
- stac-api
- raster-api
- ingest-api
- s3 stac browser for s3 origin for CF

## Acceptance Criteria

- [x] veda-backend's API Gateway values are configured to be passed down to veda-routes
- [x] veda-backend's S3 STAC Browser is configured to be passed down to veda-routes

ciaransweet:
@botanical could you define what you mean by 'propagate'? And what values from those services are being propagated? URLs?

botanical:
> @botanical could you define what you mean by 'propagate'? And what values from those services are being propagated? URLs?

@ciaransweet, definitely! I added some links in the issue description, hopefully that's helpful. In our `veda-deploy-pipeline`, I would like each service's job run to pass their origin values (potentially to `$GITHUB_OUTPUT` or some useful equivalent) so that, eventually, when we integrate `veda-routes` into the deploy pipeline, it can use those values to set the origin values in the CloudFront distribution. So, in the case of `veda-backend`, it will be passing the values for the stac-api, raster-api, ingest-api (which are all API Gateway values) and the S3 STAC browser. For more context, we have a [CDK construct](https://github.com/NASA-IMPACT/veda-backend/blob/develop/routes/infrastructure/construct.py) `veda-backend` which creates the distribution, adding origins using the env vars it has access to in `veda-backend`.

ciaransweet:
@botanical Gotcha.

So in terms of the specific values of the origins, do you mean the domains? à la: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_cloudfront_origins/HttpOrigin.html which will return a https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_cloudfront/OriginBindConfig.html#aws_cdk.aws_cloudfront.OriginBindConfig which will have a: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_cloudfront/CfnDistribution.html#aws_cdk.aws_cloudfront.CfnDistribution.OriginProperty which contains the final domain?

botanical:
> @botanical Gotcha.
> 
[...]

PR #411, opened 2024-08-02
feat: deterministic outputs of service urls

ciaransweet:
Towards #402 

# What this PR is

This PR ensures that outputs from CDK for the following services: `stac-api`, `ingest-api`, `raster-api` and `stac-browser` are available and deterministically named.

Prior to this change, if you piped the outputs to a file, you get a combination of `export_name` and a hash, which will cause issues when trying to retrieve values from the outputs for later jobs.

To get around this, I've specified the `key` parameter which explicitly tells CDK to save the outputs under a name without any logical ID logic from CDK.

I've also renamed the outputs file for this action to be `veda-backend-cdk-outputs.json` just so it's clear which stack it's for and prevents any clashes from further stacks we deploy.

# How you can test it

You can run a `cdk deploy/diff` locally and notice that the outputs are generated with the deterministic names, you can see the difference when compared to outputs I've not changed yet:

```
{
  "veda-backend-ciaran-dev": {
    "rasterapiurl": "https://333omprhn9.execute-api.us-west-2.amazonaws.com/",
    "rasterapirasterapiarn1578467D": "arn:aws:lambda:us-west-2:853558080719:function:veda-backend-ciaran-dev-rasterapilambda27FDBBDF-x0eDtlf33IGK",
    "ingestapiurl": "https://og1qrf4cu6.execute-api.us-west-2.amazonaws.com/",
    "stacapiurl": "https://9cwrbb8l4i.execute-api.us-west-2.amazonaws.com/",
    "stacbrowserurl": "https://veda-ciaran-dev-stac-browser.s3-website-us-west-2.amazonaws.com",
    "databasepgstacsecretnameBAF265DE": "a-value",
    "networkvpcid3E9D121F": "a-value"
  }
}
```

You can see that the arns are still logical output id's and so is the secret name.

### DIFF
--- a/.github/actions/cdk-deploy/action.yml
+++ b/.github/actions/cdk-deploy/action.yml
@@ -100,4 +100,4 @@ runs:
       working-directory: ${{ inputs.dir }}
       run: |
         echo $STAGE
-        cdk deploy --require-approval never --outputs-file ${HOME}/cdk-outputs.json
+        cdk deploy --require-approval never --outputs-file ${HOME}/veda-backend-cdk-outputs.json
--- a/.github/actions/cdk-deploy/action.yml
+++ b/.github/actions/cdk-deploy/action.yml
@@ -100,4 +100,4 @@ runs:
       working-directory: ${{ inputs.dir }}
       run: |
         echo $STAGE
-        cdk deploy --require-approval never --outputs-file ${HOME}/veda-backend-cdk-outputs.json
+        cdk deploy --require-approval never --outputs-file ${HOME}/cdk-outputs.json
--- a/ingest_api/infrastructure/construct.py
+++ b/ingest_api/infrastructure/construct.py
@@ -104,7 +104,7 @@ class ApiConstruct(Construct):
             "stac-ingestor-api-url",
             export_name=f"{stack_name}-stac-ingestor-api-url",
             value=self.api.url,
-            key="ingestapiurl"
+            key="ingestapiurl",
         )
 
         register_ssm_parameter(
--- a/ingest_api/infrastructure/construct.py
+++ b/ingest_api/infrastructure/construct.py
@@ -98,13 +98,13 @@ class ApiConstruct(Construct):
             custom_host=config.custom_host,
         )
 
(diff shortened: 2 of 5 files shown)


ingest_api/infrastructure/construct.py
```diff
@@ -98,13 +98,13 @@ def __init__(
             custom_host=config.custom_host,
         )
 
-        # CfnOutput(self, "ingest-api", value=self.api.url)
```
botanical, line 101: thank you for the 🧹

botanical (approved):

.github/actions/cdk-deploy/action.yml
```diff
@@ -100,4 +100,4 @@ runs:
       working-directory: ${{ inputs.dir }}
       run: |
         echo $STAGE
-        cdk deploy --require-approval never --outputs-file ${HOME}/cdk-outputs.json
+        cdk deploy --require-approval never --outputs-file ${HOME}/veda-backend-cdk-outputs.json
```
anayeaye, line 103: we're holding this PR until we have the companion change coordinated in veda-deploy, right? I think this is worth a quick shout out in tag up to get the info along to the ghg team as well.

botanical: Yes, and pending full testing.

> I think this is worth a quick shout out in tag up to get the info along to the ghg team as well.

good idea!

ciaransweet: Could you explain why this change needs holding? Is there a project expecting the `cdk-outputs.json` file down the line that would break with the rename?

Unless these renames break something, holding seems confusing to me

ciaransweet: Re-renamed back to `cdk-outputs.json`

anayeaye (approved):

ciaransweet:
Given there's no change for downstream services now, with the re-rename, is this good to merge @anayeaye @botanical ?

ciaransweet:
Looks like this can go in @anayeaye @botanical - Ran OK in the job https://github.com/NASA-IMPACT/veda-deploy/actions/runs/10280359453

PR #418, opened 2024-08-08
fix: update domain names to support multiple domain names for prod

botanical:
### What?

- Change supports multiple alternate domain names (CNAMES) for cloudfront distribution for production since it also needs an alternate domain name of `openveda.cloud`

### Why?

- So that we don't have to manually configure this and to empower those without MCP access to successfully deploy mcp-prod

### Testing?

- Tested in UAH account on new stack `veda-jt-test`


anayeaye (approved):

botanical:
Let's merge this in a separate release from the transactions work

Issue #403, opened 2024-07-22
Remove management of CloudFront distribution from veda-backend


botanical:
## Context

- Epic Link: https://github.com/NASA-IMPACT/veda-architecture/issues/469
- ADR: https://github.com/NASA-IMPACT/veda-architecture/pull/464

[veda-routes ](https://github.com/NASA-IMPACT/veda-routes) will be responsible for management CloudFront resources, so we should remove the creation of the distribution from this repo. 

## Acceptance Criteria
- [x] repo is updated to no longer create a CloudFront distribution and related resources

botanical:
@anayeaye @smohiudd , do you think this ticket should also remove the CDK for the custom API domain?

anayeaye:
@botanical you mean like `staging-stac@domain_name.com` and `staging-raster@domain_name.com`?

I think that would make the project easer to understand but I would probably only include that work in the same task if it makes it easier to remove the cloudfront distribution. I think removing that custom subdomain work might be a little messy to extract?

botanical:
> @botanical you mean like `staging-stac@domain_name.com` and `staging-raster@domain_name.com`?
> 
> I think that would make the project easer to understand but I would probably only include that work in the same task if it makes it easier to remove the cloudfront distribution. I think removing that custom subdomain work might be a little messy to extract?

@anayeaye Yep, that's what I meant! Yeah, I will take a look, I think you're right that it might be a bit messy to extract but if not, I'll include that in this issue. Thanks!

botanical:
https://github.com/NASA-IMPACT/veda-backend/pull/422

Issue #221, opened 2023-09-05
Chore: remove alt-domain support


anayeaye:
## What 
Remove configuration and support for hosting duplicate stac and raster apis at alternative domains.

## Why
The alt domain config and construct supported a one-off operational need for hosting multiple copies of the stac and raster apis at different custom domain urls. Removing this fix that is no longer needed or in use is a clean up chore only to make the project more readable.

## AC
- [ ] alt domain support removed from config and app construct

anayeaye:
https://github.com/NASA-IMPACT/veda-backend/pull/422 is going to close this issue

botanical:
https://github.com/NASA-IMPACT/veda-backend/commit/d6aedd75b9751b35c980e85ffc305933e0a87681

PR #422, opened 2024-08-15
feat: remove cloudfront distribution and custom alternate domain from backend

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/403
https://github.com/NASA-IMPACT/veda-backend/issues/221

### What + Why

- Remove cloudfront distribution CDK from veda-backend because it's now managed in veda-routes
- Update values passed into ingestor config to derive endpoint values using the custom host and stac and raster API root paths
- Remove configuration and support for hosting duplicate stac and raster apis as alternative domains.

### Testing?

Deployed to UAH


config.py
```diff
@@ -90,7 +90,12 @@ class vedaAppSettings(BaseSettings):
 
     veda_stac_root_path: str = Field(
         "",
-        description="Optional path prefix to add to all api endpoints. Used to infer url of stac-api before app synthesis.",
+        description="STAC API root path. Used to infer url of stac-api before app synthesis.",
+    )
+
+    veda_raster_root_path: str = Field(
+        "",
+        description="Raster API root path",
```
ciaransweet, line 98: Is this description the same a `veda_stac_root_path`? Or is this one not inferred

ciaransweet (approved):
Just one open question but LGTM!

config.py
botanical: I think technically it is inferred since I will be using it to produce the raster api url, but the inferences is not for the use of the stac-api before synthesis.

```
raster_api_url = f"{base_api_url}{veda_app_settings.veda_raster_root_path}/"
```

I can add that bit to the description for the raster root path if you think that's necessary!

ciaransweet: Nah all good! Was just checking!

botanical:
Going to hold off on merging this until after next release

app.py
```diff
@@ -120,23 +104,27 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
 db_secret_name = database.pgstac.secret.secret_name
 db_security_group = database.db_security_group
 
+base_api_url = f"https://{veda_app_settings.stage_name()}.{veda_app_settings.veda_custom_host}".strip(
```
smohiudd, line 107: Just a note that we're passing something similar to this `base_api_url` in both the raster and stac apis via the `VEDA_CUSTOM_HOST` env for parameter mapping: https://github.com/NASA-IMPACT/veda-backend/blob/72f446ae95ae67a57dd41dbc5baca59d02a3cfa0/stac_api/infrastructure/construct.py#L102

`VEDA_CUSTOM_HOST` was always tricky to keep track of because you need to be mindful of the stage name in the url. We could potentially remove this env and pass `base_api_url` to the stac and raster APIs instead.

smohiudd, line 107: Doesn't `VEDA_CUSTOM_HOST` already include the stage name?

app.py
```diff
@@ -120,23 +104,27 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+stac_api_url = f"{base_api_url}{veda_app_settings.veda_stac_root_path}/"
+raster_api_url = f"{base_api_url}{veda_app_settings.veda_raster_root_path}/"
+
 # ingestor config requires references to other resources, but can be shared between ingest api and bulk ingestor
 ingestor_config = ingest_config(
     stage=veda_app_settings.stage_name(),
     stac_db_security_group_id=db_security_group.security_group_id,
-    stac_api_url=stac_api.stac_api.url,
-    raster_api_url=raster_api.raster_api.url,
+    stac_api_url=stac_api_url,
```
anayeaye, line 117: I think we could handle inferring the cloudfront urls in config with a property or function
```
@property
def veda_stac_api_cf_url(self) -> str:
    """inferred cloudfront url of the stac api if app is configured with a custom host and root path"""
    if self.veda_custom_host and self.veda_stac_root_path:
        return f"https://{self.veda_custom_host}{self.veda_stac_root_path}"
    return None 
```

and then conditionally use them here so that we have a way to use the API gateway url if a cloudfront is not used. 

```
stac_api_url=settings.stac_api_cf_url if settings.stac_api_cf_url else stac_api.stac_api.url
```

app.py
botanical: Ah it does!

botanical: https://github.com/NASA-IMPACT/veda-backend/pull/422/commits/f0e088488e59f94207e427822f127965a0a5db11

smohiudd (approved):

anayeaye (commented):
Sorry I didn't hit send on these yesterday. The only change I think we might need is the suggestion about handling both api gateway urls and cloudfront urls https://github.com/NASA-IMPACT/veda-backend/pull/422#discussion_r1727340610

ingest_api/infrastructure/config.py
```diff
@@ -87,3 +95,17 @@ def env(self) -> aws_cdk.Environment:
             account=self.aws_account,
             region=self.aws_region,
         )
+
+    @property
```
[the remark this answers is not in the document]
botanical: I think I would prefer to keep it in the ingest_api for now since, in the future, we want to block access via apigw anyway, right? 🤔 I'm open to moving it though!

ingest_api/infrastructure/construct.py
```diff
@@ -50,11 +45,11 @@ def __init__(
             "DYNAMODB_TABLE": self.table.table_name,
             "JWKS_URL": self.jwks_url,
             "NO_PYDANTIC_SSM_SETTINGS": "1",
-            "STAC_URL": config.stac_api_url,
+            "STAC_URL": config.veda_stac_api_cf_url,
```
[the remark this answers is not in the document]
botanical: Oh! Sorry I misunderstood - https://github.com/NASA-IMPACT/veda-backend/pull/422/commits/8735f45905df21b60a75ced438b273496faf762d

ingest_api/infrastructure/config.py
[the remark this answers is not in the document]
anayeaye: Let's keep it here then 👍

anayeaye (approved):

PR #431, opened 2024-09-24
fix: update stac_api_url and raster_api_url properties to check correct variables

botanical:
### What? / Why?

- Update the condition to check the correct variable

### Testing?

- Deployed to UAH dev for testing


anayeaye (approved):
Should we save this merge to trigger our dev database restore?

anayeaye (approved):

anayeaye (approved):

PR #439, opened 2024-10-14
ci: fix cicd target branch name and fix release workflow

anayeaye:
## what
- Fix: correct the name of the target branch for the develop deployment cicd
- Fix: return the release workflow trigger for pushes to develop and main branches

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -7,7 +7,7 @@ permissions:
 on:
   push:
     branches:
-      - dev
+      - develop
 
 jobs:
   define-environment:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
+name: Release Workflow
+
+permissions:
+  id-token: write
+  contents: read
```
botanical, line 5: what are these top level permissions for?

.github/workflows/cicd.yml
```diff
@@ -7,7 +7,7 @@ permissions:
 on:
   push:
     branches:
-      - dev
+      - develop
```
botanical, line 10: 👍

botanical (approved):

.github/workflows/release.yml
anayeaye: I think this is needed for the tagging

PR #446, opened 2024-10-28
feat(stac)!: upgrade to stac-fastapi-pgstac

anayeaye:
### Issue

This PR is just a test to see what the cdk diff looks like on a github runner


anayeaye:
Done testing dependency changes.

Issue #437, opened 2024-10-07
Upgrade stac-fastapi >=3.0


smohiudd:
## Description

The STAC Api is using stac-fastapi v2.4 which uses Pydantic v1. stac-fastapi >=3.0 uses Pydantic v2

## Acceptance Criteria
- [x] STAC API (stac-fastapi) is using >=3.0

Issue #436, opened 2024-10-07
Upgrade stac-pydantic to use pydantic v2


smohiudd:
## Description

The Ingest API is using a [forked/pinned version](https://github.com/NASA-IMPACT/veda-backend/blob/72f446ae95ae67a57dd41dbc5baca59d02a3cfa0/ingest_api/runtime/requirements.txt#L14) of stac-pydantic which is still using pydantic v1. The changes from that fork should now be available in a new version of stac-pydantic so we should now upgrade to stac-pydantic >=v3.0.x. (which uses pydantic v2).

## Acceptance Criteria
- [x] Ingest API (stac-pydantic) is using stac-pydantic version that uses pydantic v2 and tested using endpoints and Airflow

PR #449, opened 2024-12-19
fix: Pydantic, Pypgstac and Python Upgrades

ividito:
### Issue

#435 
#436 
#437 

### What?

- Upgraded Pydantic to version 2, which required several changes to validators. In particular, most decorators changed, and all `Optional` values must explicitly default to `None`.
- Upgraded pypgstac to 0.8.5
- Upgraded Python runtimes to avoid errors installing Pydantic on earlier Python versions

### Testing?

- Tested manually locally via docker compose, and tested CDK build.


ingest_api/runtime/tests/conftest.py
```diff
@@ -114,18 +114,21 @@ def example_stac_item():
         "links": [
             {
                 "rel": "collection",
+                "label": None,
```
anayeaye, line 117: I think 'label' is allowed but not actually a property in the spec https://github.com/radiantearth/stac-api-spec/blob/release/v1.0.0/stac-spec/collection-spec/collection-spec.md#link-object

Or is this related to something else?

ingest_api/runtime/tests/test_registration.py
```diff
@@ -32,32 +32,37 @@ def populate_table(self, count=100) -> List["schemas.Ingestion"]:
         for i in range(count):
             ingestion = self.example_ingestion.copy()
             ingestion.id = str(i)
-            ingestion.created_at = ingestion.created_at + timedelta(hours=i)
+            ingestion.created_at = datetime.now() + timedelta(hours=i)
```
anayeaye, line 35:
```suggestion
            ingestion.created_at = datetime.utcnow() + timedelta(hours=i)
```

stac_api/infrastructure/config.py
```diff
@@ -45,12 +46,13 @@ class vedaSTACSettings(BaseSettings):
     )
 
     userpool_id: Optional[str] = Field(
-        description="The Cognito Userpool used for authentication"
+        "None", description="The Cognito Userpool used for authentication"
```
anayeaye, line 49:
```suggestion
        None, description="The Cognito Userpool used for authentication"
```
Because this is cast as string in the construct does it need to be a string default?

anayeaye (commented):
Just a few small comments/questions. Also just a reminder: we should update our secret for the dev instance with the new pgstac version, too. I'm excited to see this in dev!

stac_api/infrastructure/config.py
ividito: Nope, but it needs to have an explicit default, this is meant to just be `None`. Good catch!

ingest_api/runtime/tests/conftest.py
ividito: I think when pytest pulls the model in it autopopulates a `None`, so the assertion was failing. This could be a `stac_pydantic.Item` bug based on that spec.

anayeaye: Gotcha https://github.com/stac-utils/stac-pydantic/blob/main/stac_pydantic/links.py#L22

anayeaye (approved):

smohiudd:
Local tests worked for me. This looks good!

smohiudd (approved):

PR #451, opened 2025-01-07
fix: Change stac_extension URL serialization

ividito:
Pydantic v2 changes URL type behaviour, preventing JSON serialization (which we use to load collections into our catalog). We removed most instances of these types when we migrated to v2, but the stac-pydantic `Catalog` model (used in the `Collection` model) still uses `AnyUrl` for `stac_extensions`.

I'm not convinced this is necessarily a bug, but it is blocking our ingests, so I'm including a fix here rather than upstream (for now). There are changes to URLs on the pydantic v3 roadmap, which may or may not fix this.

Ref:
https://github.com/pydantic/pydantic/discussions/8211
https://github.com/pydantic/pydantic/issues/7186

### DIFF
--- a/ingest_api/runtime/src/schemas.py
+++ b/ingest_api/runtime/src/schemas.py
@@ -8,12 +8,14 @@ from urllib.parse import urlparse
 
 import src.validators as validators
 from pydantic import (
+    AnyUrl,
     BaseModel,
     ConfigDict,
     Field,
     Json,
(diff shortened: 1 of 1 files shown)


anayeaye (approved):
Good find!
Also: https://github.com/NASA-IMPACT/veda-backend/issues/452 😂

PR #454, opened 2025-01-09
fix: change dynamo db method reference

smohiudd:
Fix the dynamo table write method to use the correct ingestion schema method name:

```
def write(self, ingestion: schemas.Ingestion):
        self.table.put_item(Item=ingestion.dynamodb_dict())
```


botanical (approved):
Tested in dev

PR #467, opened 2025-03-12
feat: add eoapi-auth-util to stac api

smohiudd:
- Add eoapi-auth-utils to stac api
- Add new envs `VEDA_KEYCLOAK_CLIENT_ID` and `VEDA_OPENID_CONFIGURATION_URL`
- remove Cognito variables from STAC API

### Testing?

- TBD


smohiudd:
Authorized in the STAC *without being added to a group* and tried to post a new collection to `/collections` endpoint. Received a 401 error and the following message.
`
{
  "detail": "Could not validate credentials"
}
`

I believe this is expected since I'm not part of the editor or admin groups in keycloak.

smohiudd:
The above error was not because I wasn't  part of a group but instead because `allowed_jwt_audiences` was defaulting to None in `OpenIdConnectAuth`. Setting to the "account" aud fixed the problem as I was able to ingest a collection using the /collections POST endpoint.

stac_api/runtime/src/config.py
```diff
@@ -63,55 +63,14 @@ class _ApiSettings(BaseSettings):
...
-        "", description="The Cognito APP client secret"
+    client_id: Optional[str] = Field(None, description="Auth client ID")
+    openid_configuration_url: Optional[AnyHttpUrl] = Field(
+        None, description="OpenID config url"
     )
     enable_transactions: bool = Field(
         False, description="Whether to enable transactions"
     )
 
-    @model_validator(mode="before")
```
smohiudd, line 82: This model validator is already included in `/infrastructure/config` so this would be a duplicate validator.

j08lue:
For the record - this contributes to the Auth goal of integrating auth into STAC API

- https://github.com/NASA-IMPACT/veda-keycloak/issues/49

botanical (commented):
Thanks for your assistance in answering my questions on how to test this @smohiudd! From my testing, this changes seems to function but I did notice that when I go through the flow of authorizing via cilogon, I get redirected to a blank site and then I have to manually return the the STAC api docs page and refresh and then it shows that I'm authorized.

smohiudd:
Thanks @botanical for testing this! That redirect to blank screen is expected and something we'll need to work on fixing at a later time.

botanical (approved):
Talked during tag up and these changes look good

PR #468, opened 2025-03-14
fix: add default for collection assets

smohiudd:
Add default for assets in ingest api collection validation so it's not a required field in pydantic  v2 as per the migration guide:
https://docs.pydantic.dev/dev/migration/#required-optional-and-nullable-fields


ingest_api/runtime/src/schemas.py
```diff
@@ -66,7 +66,7 @@ class DashboardCollection(Collection):
     time_density: Optional[str] = Field(default=None, alias="dashboard:time_density")
     item_assets: Optional[Dict]
     links: Optional[List[LinkWithExtraFields]]
-    assets: Optional[Dict]
```
botanical, line 69: Thanks!

botanical (approved):

PR #481, opened 2025-05-06
fix: remove allow public subnets from lambdas

smohiudd:
- remove `allow_public_subnet` from stac and raster api since they use private subnets
- remove unused `db_vpc_subnets` paramater in ingest api
- remove `allow_public_subnet` from `IngestorConstruct` lambda since it should be placed in private subnet and needs internet access for keycloak auth

## Testing

- This branch was deployed to test.openveda.cloud and [ida-ndwi-difference-test-ingest-api-keycloak](https://dev.openveda.cloud/stac/collections/ida-ndwi-difference-test-ingest-api-keycloak) was ingested using the ingest API directly

### DIFF
--- a/ingest_api/infrastructure/construct.py
+++ b/ingest_api/infrastructure/construct.py
@@ -24,7 +24,6 @@ class ApiConstruct(Construct):
         config: IngestorConfig,
         db_secret: secretsmanager.ISecret,
         db_vpc: ec2.IVpc,
-        db_vpc_subnets=ec2.SubnetSelection,
         **kwargs,
     ) -> None:
         super().__init__(scope, construct_id, **kwargs)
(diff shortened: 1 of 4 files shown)


ingest_api/infrastructure/construct.py
```diff
@@ -24,7 +24,6 @@ def __init__(
         config: IngestorConfig,
         db_secret: secretsmanager.ISecret,
         db_vpc: ec2.IVpc,
-        db_vpc_subnets=ec2.SubnetSelection,
```
smohiudd, line 27: `db_vpc_subnets` parameter is not used in this class

ingest_api/infrastructure/construct.py
```diff
@@ -285,8 +281,6 @@ def build_ingestor(
             timeout=Duration.seconds(180),
             environment={"DB_SECRET_ARN": db_secret.secret_arn, **env},
             vpc=db_vpc,
-            vpc_subnets=db_vpc_subnets,
```
smohiudd, line 288: This lambda does not need to be place in a public subnet

anayeaye (approved):

Issue #482, opened 2025-05-08
Upgrade titiler and rio-tiler


smohiudd:
## Description

Upgrade titiler to 0.19 ([breaking changes](https://github.com/developmentseed/titiler/blob/main/CHANGES.md#0190-2024-11-07)). This include an upgrade to rio-tiler dependency to `>=7.0,<8.0`

PR #483, opened 2025-06-09
fix(dependencies): titiler and rio-tiler upgrade

ividito:
### Issue

#482 

### What?

- Updates raster API runtime to 3.12
- Updates titiler and titiler-pgstac
- Updates tests to use new endpoints for breaking changes to point queries and TMS inputs
- Flag to disable tests in CI (In veda-deploy, we perform integration tests against deployed services as a separate step after deployment. Disabling the pre-deploy tests makes those deployments faster. Default CI behaviour is unchanged.)
~- [ ] TODO some endpoints no longer have debug headers, I would like to re-enable these before merging, but this is fairly low priority~ see comments

### Testing?

- Pending some fixes to our SIT environment, this has been deployed to https://ig3cl55i3b.execute-api.us-west-2.amazonaws.com/docs
- Integration+unit tests pass
- Local manual tests pass/seem fine
- In particular, I reproduced the issue documented in https://github.com/US-GHG-Center/ghgc-architecture/issues/220 to make sure that we don't see regression in overview quality/accuracy

### DIFF
--- a/.github/workflows/tests/conftest.py
+++ b/.github/workflows/tests/conftest.py
@@ -18,6 +18,7 @@ SEARCH_ROUTE = "search"
 RASTER_SEARCHES_ENDPOINT = "http://0.0.0.0:8082/searches"
 RASTER_HEALTH_ENDPOINT = "http://0.0.0.0:8082/healthz"
 TILEMATRIX = {"z": 15, "x": 8589, "y": 12849}
+TMS_ID = "WebMercatorQuad"
 
 SEARCHES = [
     {
@@ -226,6 +227,17 @@ def seeded_tilematrix():
         dict: A [z, x, y] set of dimensions
     """
     return TILEMATRIX
+
+
+@pytest.fixture
+def seeded_tms_id():
+    """
+    Fixture providing a matrix of seeded data for integration testing.
+
+    Returns:
+        dict: A [z, x, y] set of dimensions
+    """
+    return TMS_ID
 
 
 @pytest.fixture
--- a/.github/workflows/tests/conftest.py
+++ b/.github/workflows/tests/conftest.py
@@ -235,7 +235,7 @@ def seeded_tms_id():
     Fixture providing a matrix of seeded data for integration testing.
 
     Returns:
-        dict: A [z, x, y] set of dimensions
+        string: A valid TileMatrixSet ID
     """
     return TMS_ID
 
--- a/raster_api/runtime/src/app.py
+++ b/raster_api/runtime/src/app.py
@@ -23,7 +23,6 @@ from titiler.core.factory import (
     TMSFactory,
 )
 from titiler.core.middleware import CacheControlMiddleware
-from titiler.core.resources.enums import OptionalHeader
 from titiler.core.resources.responses import JSONResponse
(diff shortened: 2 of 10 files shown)


.github/workflows/tests/conftest.py
```diff
@@ -228,6 +229,17 @@ def seeded_tilematrix():
     return TILEMATRIX
 
 
+@pytest.fixture
+def seeded_tms_id():
+    """
+    Fixture providing a matrix of seeded data for integration testing.
+
+    Returns:
+        dict: A [z, x, y] set of dimensions
```
anayeaye, line 238:
```suggestion
        string: A valid TileMatrixSet ID
```

raster_api/runtime/src/app.py
```diff
@@ -148,14 +148,14 @@ async def lifespan(app: FastAPI):
...
     path_dependency=ItemIdParams,
-    optional_headers=optional_headers,
     router_prefix="/collections/{collection_id}/items/{item_id}",
     environment_dependency=settings.get_gdal_config,
     router=APIRouter(route_class=LoggerRouteHandler),
     extensions=[
         stacViewerExtension(),
     ],
     colormap_dependency=ColorMapParams,
+    # optional_headers=optional_headers,
```
anayeaye, line 158: I can't find the removal of optional_headers from `TilerFactory` in the titiler changelog. Did you discover it isn't working and that is why it is supressed here? Or did have we been supplying unused optional_headers all along?
Maybe a quick note on these suppressed lines about why we are not passing this param?

anayeaye:
I have been evaluating the changes with this dashboard preview https://deploy-preview-875--visex.netlify.app/ and all seems to be working as expected for exploration and analysis and the stories I skimmed.

As for the MiCASA issue: it seems like it was resolved by [increasing the data resolution](https://github.com/US-GHG-Center/ghgc-architecture/issues/220#issuecomment-2430154918) so perhaps it is no longer an issue?

I did observe some flickering behavior as the very fine pixels are interpolated at increasing resolutions on zooming some of the less-standard-formatting datasets like facebook population density but this behavior is also present in the production dashboard.

raster_api/runtime/src/app.py
anayeaye: > TODO some endpoints no longer have debug headers, I would like to re-enable these before merging, but this is fairly low priority
Or can we just re-enable them? Not sure why we removed them

ividito: I ran into this error on startup with optional headers included on all factories.
```
veda.raster    |   File "/opt/bitnami/python/lib/python3.12/site-packages/src/app.py", line 148, in <module>
veda.raster    |     stac = MultiBaseTilerFactory(
veda.raster    |            ^^^^^^^^^^^^^^^^^^^^^^
veda.raster    | TypeError: MultiBaseTilerFactory.__init__() got an unexpected keyword argument 'optional_headers'
```

The error stems from this diff in titiler: https://github.com/developmentseed/titiler/pull/943 - I think this just means we can't use the optional headers outside of the Mosaic tiler, unless we directly add support for them. I'm not sure that this is necessary - I've used the server timing headers in the past, but I've switched to monitoring Lambda response times.

Issue #494, opened 2025-07-03
Fix: include tileMatrixSetId in STAC Item map links [Impacted by v13.0.0-patch.1 decisions]


anayeaye:
## What
Update dynamically rendered titiler map links to include required. tile matrix set id.

## Notes
```
href = urljoin(
            self.tiler_href, f"collections/{collection_id}/items/{item_id}/map?{qs}"
        )
```

should be
```
href = urljoin(
            self.tiler_href, f"collections/{collection_id}/items/{item_id}/WebMercatorQuad/map?{qs}"
        )
```

in stac_api/runtime/src/links.py https://github.com/NASA-IMPACT/veda-backend/blob/v13.0.0-patch.1/stac_api/runtime/src/links.py#L54

## Titiler core changelog
https://github.com/developmentseed/titiler/blob/main/CHANGES.md#misc-4

## AC
- [ ] item map links work
- [ ] item preview links work (when this ticket was created, previews worked as they are but may later require a tileMatrixSetId parameter

anayeaye:
Dynamic Map of Item links are working in this example https://dev.openveda.cloud/api/stac/collections/hd-blackmarble-nightlights/items/hdnightlights_2023-03-15_32_97N90_96W32_83N90_79W-day

https://dev.openveda.cloud/api/raster/collections/hd-blackmarble-nightlights/items/hdnightlights_2023-03-15_32_97N90_96W32_83N90_79W-day/WebMercatorQuad/map?bidx=1&bidx=2&bidx=3&assets=cog_default&rescale=0%2C255 

<img width="1375" height="934" alt="Image" src="https://github.com/user-attachments/assets/2405f74b-3d77-4780-85ac-b275f65adb1f" />

PR #502, opened 2025-07-23
fix: include tileMatrixSetId in STAC Item map links

stephenkilbourn:
### Issue

#494 

### What?

Update dynamically rendered titiler map links to include required tile matrix set id.

### Why?

to accomodate changes in[ titiler v0.19.0](https://github.com/developmentseed/titiler/blob/main/CHANGES.md#0190-2024-11-07) 

### Testing?

- Relevant testing details


stephenkilbourn:
i don't think that it is an urgent thing to deploy right now

PR #526, opened 2025-09-15
feat: switch to improved monitoring middleware

ividito:
### Issue

https://github.com/NASA-IMPACT/veda-architecture/issues/655

### What?

- Switch to using a middleware for logging, metrics, and traces (which are still limited until we get newer versions of titiler deployed)
- Adjust metrics to have more meaningful dimensions

### Why?

- Consistency with other services
- Improved metrics to support meaningful alarms and dashboards

### Testing?

- Currently deployed to SIT
- New metrics currently available on SIT and dev monitoring stacks - see WIP dashboard on https://monitoring.dev.openveda.cloud/

### DIFF
--- a/ingest_api/infrastructure/config.py
+++ b/ingest_api/infrastructure/config.py
@@ -1,3 +1,4 @@
+import subprocess
 from getpass import getuser
 from typing import List, Optional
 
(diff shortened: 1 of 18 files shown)


stephenkilbourn (approved):

ingest_api/infrastructure/config.py
```diff
@@ -93,6 +94,11 @@ class IngestorConfig(BaseSettings):
         case_sensitive=False, env_file=".env", env_prefix="VEDA_", extra="ignore"
     )
 
+    git_sha: Optional[str] = Field(
+        subprocess.check_output(["git", "rev-parse", "HEAD"]).strip().decode("utf-8"),
```
botanical, line 98: do we want to decode first (convert bytes to string) and then remove whitespace? https://docs.python.org/3/library/subprocess.html#subprocess.TimeoutExpired.output 
```suggestion
        subprocess.check_output(["git", "rev-parse", "HEAD"]).decode("utf-8").strip(),
```

PR #531, opened 2025-09-19
fix: experimenting with stac auth proxy configure_app

botanical:
### Issue

Link to relevant GitHub issue

### What?

- Description of the changes made

### Why?

- Description of why the changes were made

### Testing?

- Relevant testing details

PR #535, opened 2025-10-09
fix: tenant filtering via stac auth proxy

alukach:
### What?

This PR represents recommendations of updates to #531 to better integrate with the stac-auth-proxy.

### How?

The primary adjustments made were:

1. Remove any attempt at applying tenant filters to the incoming requests. This is the responsibility of the filtering tooling that is built into the stac-auth-proxy. Instead, the STAC API should provide filter generators[^1] that build the CQL2 filters, trusting that the stac-auth-proxy will apply them to the items and collections endpoints.
2. Break-up tenant middleware into two separate middleware for simplicity and better encapsulation
3. Move tenant middleware to _after_ the application of the stac-auth-proxy, ensuring that it runs _before_ the stac-auth-proxy's middleware. This way, the tenant is extracted and the `path` is cleaned up before the stac-auth-proxy processes requests.
2. Remove compression from `stac-auth-proxy`, perform it after all middleware runs (this is a reversal of my previous suggestion, sorry 🤕, I wasn't aware that we would need to process links after the stac-auth-proxy's middleware ran)

> [!IMPORTANT]
> There are some issues with the current queries (as documented within the `ItemFilter` class) preventing us from querying across multiple collections. We'll need to sync with @bitner to assess the cause.

#### Along the way...

- Docker Compose file
  - Reduce the concurrency of the STAC API to just a single worker. I think this reduces visual noise when running docker compose locally.
  - Adding a dummy OIDC server for testing when we do want to start making use of auth
  - Specifying the `root_path` to better match production
  - Fix env var to enable stac-auth-proxy

### Testing?

1. Run dev env:
    ```
    docker compose up
    ```
2. Run through scenarios:
    ```
    # Successfully return unfiltered collections without tenant
    ▶ curl -s http://localhost:8081/api/stac/collections | jq .numberReturned                              
    1
    
    # Successfully return collection items without a tenant
    ▶ curl -s http://localhost:8081/api/stac/collections/noaa-emergency-response/items | jq .numberReturned    
    10
    
    # Successfully return unfiltered search without tenant
    ▶ curl -s http://localhost:8081/api/stac/search | jq .numberReturned     
    10
    
    # Successfully return collections with a tenant
    ▶ curl -s http://localhost:8081/api/stac/fake-tenant/collections | jq .numberReturned                              
    1
    
    # Successfully return collection items with a tenant
    ▶ curl -s http://localhost:8081/api/stac/fake-tenant/collections/noaa-emergency-response/items | jq .numberReturned
    10
    
    # Successfully return search with a tenant
    ▶ curl -s http://localhost:8081/api/stac/fake-tenant/search | jq .numberReturned
    10
    
    # Successfully return search with a tenant
    ▶ curl -s http://localhost:8081/api/stac/fake-tenant/search | jq .numberReturned
    10
    
    # Return no collections for unknown tenant
    ▶ curl -s http://localhost:8081/api/stac/foo/collections | jq .numberReturned        
    0
    
    # Return no collection items for unknown tenant
    ▶ curl -s http://localhost:8081/api/stac/bar/collections/noaa-emergency-response/items | jq .numberReturned        
    0
    
    # Return no search for unknown tenant
    ▶ curl -s http://localhost:8081/api/stac/xyz/search | jq .numberReturned        
    0

    # Correctly places tenant in search links
    ▶ curl -s http://localhost:8081/api/stac/fake-tenant/search | jq '.features[0].links[0]'
    {
      "rel": "collection",
      "type": "application/json",
      "href": "http://localhost:8081/api/stac/fake-tenant/collections/noaa-emergency-response"
    }
    
    # Correctly render links without tenant for search
    ▶ curl -s http://localhost:8081/api/stac/search | jq '.features[0].links[0]'     
    {
      "rel": "collection",
      "type": "application/json",
      "href": "http://localhost:8081/api/stac/collections/noaa-emergency-response"
    }
    
    # Correctly places tenant in collection links
    ▶ curl -s http://localhost:8081/api/stac/fake-tenant/collections | jq '.collections[0].links[0]'
    {
      "rel": "items",
      "type": "application/geo+json",
      "href": "http://localhost:8081/api/stac/fake-tenant/collections/noaa-emergency-response/items"
    }
    
    # Correctly render links without tenant for collections
    ▶ curl -s http://localhost:8081/api/stac/collections | jq '.collections[0].links[0]'            
    {
      "rel": "items",
      "type": "application/geo+json",
      "href": "http://localhost:8081/api/stac/collections/noaa-emergency-response/items"
    }

    ```

[^1]: https://developmentseed.org/stac-auth-proxy/user-guide/record-level-auth/#filter-contract


docker-compose.yml
```diff
@@ -30,18 +30,20 @@ services:
       # https://github.com/developmentseed/eoAPI/issues/16
       # - TITILER_ENDPOINT=raster
       - TITILER_ENDPOINT=http://0.0.0.0:8082
-      - VEDA_STAC_ROOT_PATH=
-      - VEDA_STAC_OPENID_CONFIGURATION_URL=https://example.com/.well-known/openid-configuration
+      - VEDA_STAC_ROOT_PATH=/api/stac
```
botanical, line 33: 👍

botanical (approved):

stac_api/runtime/src/app.py
```diff
@@ -91,14 +95,32 @@ async def lifespan(app: FastAPI):
...
-    # Use standard FastAPI app when authentication is disabled, for testing
-    # and add compression middleware since stac-auth-proxy provides it when enabled
+    # Use standard FastAPI app when authentication is disabled
     app = api.app
-    app.add_middleware(CompressionMiddleware)
+
+# Note: we want this to be added after stac_auth_proxy so that it runs before stac_auth_proxy's middleware
+app.add_middleware(TenantExtractionMiddleware, root_path=api_settings.root_path)
+app.add_middleware(TenantLinksMiddleware, root_path=api_settings.root_path)
+app.add_middleware(CompressionMiddleware)
```
botanical, line 125: Could you help me understand the final request path based on the middleware ordering? 

So if we provide a tenant in the collections path like `/api/stac/test-tenant/collections`, then TenantLinksMiddleware adds the tenant to all the links in the json response, TenantExtractionMiddleware extracts the tenant from the url, and then CQL2 filtering is applied? Do we want to extract the tenant first? @alukach

alukach: I think of middleware as a wrapper on the API. Each time you add another middleware, you wrap the API with another layer.  So the last middleware to get added will be on the outside of the application.  When a request enters the application, it goes through those layers of middleware, accesses the application in the center, and then the response goes back out of the layers of middleware in the opposite order.

So if we have our middleware applied like this:

```py
app.add_middleware(StacAuthProxyMiddleware)     # 1
app.add_middleware(TenantExtractionMiddleware)  # 2
app.add_middleware(TenantLinksMiddleware)       # 3
app.add_middleware(CompressionMiddleware)       # 4
```

The middleware will be used in order 4 to 1 on the request, and then order 1 to 4 on the response.

A crude diagram:

```mermaid
sequenceDiagram
    participant Client
    participant CompressionMiddleware
    participant TenantLinksMiddleware
    participant TenantExtractionMiddleware
    participant proxy as stac-auth-proxy
    participant App as Application

    Client->>CompressionMiddleware: Incoming Request
    CompressionMiddleware->>TenantLinksMiddleware: Pass request (unaltered)
    TenantLinksMiddleware->>TenantExtractionMiddleware: Pass request (unaltered)
    TenantExtractionMiddleware->>proxy: Store tenant, alter request path

    proxy->>App: Add filters to request (maybe)

    App-->>proxy: Respond with STAC items
    proxy-->>TenantExtractionMiddleware: Validate response against filters,<br />rewrite links,add authentication<br />extension to JSON body
    TenantExtractionMiddleware-->>TenantLinksMiddleware: Pass response (unaltered)
    TenantLinksMiddleware-->>CompressionMiddleware: Update links with tenant in URL<br /> (if tenant is present)
    CompressionMiddleware-->>Client: Response Compressed
```

All this is to say, the tenant is currently extracted first, before the stac-auth-proxy does its thing.  This is nice, because as you wrote, we can remove the tenant from the path after we read it and the stac-auth-proxy can proceed as if this were a typical single-tenant system.

docker-compose.yml
anayeaye, line 33: We will need to update our tests to expect this root path

alukach: Do the tests use docker compose?

docker-compose.yml
```diff
@@ -30,18 +30,20 @@ services:
...
+      - VEDA_STAC_ENABLE_STAC_AUTH_PROXY=True
     depends_on:
       - database
       - raster
+      - oidc
     command: bash -c "bash /tmp/scripts/wait-for-it.sh -t 120 -h database -p 5432 && /start.sh"
     volumes:
       - ./scripts:/tmp/scripts
-
+      - ./stac_api/runtime/src:/opt/bitnami/python/lib/python3.12/site-packages/src/
```
anayeaye, line 46: maybe we should update the local/Dockerfiles to use `public.ecr.aws/sam/build-python3.12:latest` instead of ghcr.io/vincentsarago/uvicorn-gunicorn:3.10 so we wont need bitnami images?

anayeaye: I'll open an issue to handle that separately

docker-compose.yml
anayeaye: Yeah some do. I think these are resolved in #535 though

PR #538, opened 2025-10-10
test: functional tenant filter tests

alukach:
### What?

- extend test fixture data to include a variety of scenarios:
  1. tenant `emergency` with single collection: `noaa-emergency-response`
  1. tenant `fire` with multiple collections: `barc-thomasfire`, `caldor-fire-behavior`
  2. collection without a tenant: `CMIP245-winter-median-pr`
- buildout tests to ensure tenant filtering operates as expected
  - Tenant collections should be listed when accessed with valid tenant.
  - All collections should be listed when accessed without tenant.
  - No collections should be listed when accessed with invalid tenant.
  - A tenant's collection should be available when accessed with that tenant.
  - Any collection should be available when accessed without tenant.
  - No collection should be returned when accessed with invalid tenant.
  - Items should be available when accessed with valid tenant.
  - All collection's items should be available when accessed without tenant.
  - No collection's items should be returned when accessed with invalid tenant.
  - Search with a tenant should return only items from that tenant's collections.
  - Search without a tenant should return all collections.
  - Search should return no items when accessed with invalid tenant.
  - Tenants should not be able to access each other's collections or items.

### Why?

Validate the functionality of #531

### Testing?

Locally:

1. `docker compose up`
2. `python -m pytest .github/workflows/tests/test_stac.py -vv -s -k TestTenantFiltering`


.github/workflows/tests/test_stac.py
```diff
@@ -156,3 +157,236 @@ def test_stac_to_raster(self):
...
+
+    def test_collections_listed_without_tenant(self):
+        """All collections should be listed when accessed without tenant."""
+        resp = httpx.get(f"{self.stac_endpoint}/{self.collections_route}")
+        assert resp.status_code == 200
+        all_collections = set(
+            collection
+            for _tenant_collections in self.tenant_collections.values()
+            for collection in _tenant_collections
+        )
```
botanical: Any reason not to use a simple loop such as:
```
all_collections = set()
for collections in self.tenant_collections.values():
    all_collections.update(collections)
```?

File: .github/workflows/tests/test_stac.py [hunk 2]
```diff
@@ -156,3 +157,236 @@ def test_stac_to_raster(self):
...
+    def test_tenant_isolation(self, tenant):
+        """Tenants should not be able to access each other's collections or items."""
+        # emergency should not see barc-thomasfire
+        resp = httpx.get(f"{self.stac_endpoint}/{tenant}/collections")
+        assert resp.status_code == 200
+        other_tenants_collections = set(
+            collection
+            for tenant, collection in self.tenant_collections.items()
+            if tenant != tenant
+        )
```
botanical:
```suggestion
        other_tenants_collections = set(
            collection_id
            for tenant_name, collections in self.tenant_collections.items()
            if tenant_name != tenant
            for collection_id in collections
        )
```
I think this change generates the sets we want to use since `if tenant != tenant` just compares it to itself, so it will always be `False`

Also, maybe it's easier to read if we used a loop instead like
```
other_tenants_collections = set()
for tenant_name, collections in self.tenant_collections.items():
    if tenant_name != tenant:
        other_tenants_collections.update(collections) 
```

botanical (changes requested):
These tests are great! There was one test that had a condition that looked iffy to me that I think we should update.

.github/workflows/tests/test_stac.py
anayeaye: ➕ 1

I ran @botanical's suggested loop and verified it is what we want locally (and that the set as written in the pr is empty
```
other_tenants_collections = set()
for tenant_name, collections in self.tenant_collections.items():
    if tenant_name != tenant:
        other_tenants_collections.update(collections) 
```

anayeaye (approved):
approve with the one suggested change from @botanical

.github/workflows/tests/test_stac.py
alukach: Yes, purely a stylistic decision. Your code looks fine and likely more legible, happy to swap it for what I had written.

botanical: Maybe it wouldn't hurt us to get used to reading generator expressions 😅  We can keep it as is

botanical (approved):

stac_api/runtime/tests/test_extensions.py
```diff
@@ -122,9 +130,9 @@ async def test_get_collection_by_id(self, api_client, collection_in_db):
         # The `collection_in_db` fixture ensures the collection exists and provides its ID.
         collection_id = collection_in_db
 
-        # Perform a GET request to the /collections endpoint with an "ids" query
+        # Perform a GET request to the /collections endpoint with an "id" query
         response = await api_client.get(
-            collections_endpoint, params={"ids": collection_id}
```
alukach, line 127: @anayeaye did a query like this ever work?  I couldn't seem to get it to respond with filtered results:

```
▶ curl -s https://openveda.cloud/api/stac/collections\?id\=bangladesh-landcover-2001-2020-test | jq .numberMatched
247
```

anayeaye: @alukach I guess not (but hadn't yet had reason to use a collection id in a GET collections request when we have `/collections/{collection_id}`)


Also 
```
curl -s "https://openveda.cloud/api/stac/collections?collections=bangladesh-landcover-2001-2020" | jq .numberMatched
247
```
🤔 

Some requests do seem to work using filter, but the syntax is weird and required month in single quotes (double doesn't work `dashboard:time_density='month'`)
```
curl -X 'GET' \                  
  'https://test.openveda.cloud/api/stac/collections?limit=10&filter=dashboard%3Atime_density%3D%27month%27&filter-lang=cql2-text' \
  -H 'accept: application/json' | jq .numberMatched

13
```

PR #540, opened 2025-10-14
feat: cache collections lookup

alukach:
### What?

Cache collection lookups in the items filter factory

### Why?

When we fetch a tenant's collection to build an items filter, we should cache that value to reduce pressure on the API

### Testing?

- Relevant testing details

### DIFF
--- a/stac_api/runtime/setup.py
+++ b/stac_api/runtime/setup.py
@@ -9,6 +9,7 @@ with open("README.md") as f:
 
 inst_reqs = [
     "boto3",
+    "cachetools>=6.2.1",
     "stac-fastapi.api~=5.0",
     "stac-fastapi.types~=5.0",
     "stac-fastapi.extensions~=5.0",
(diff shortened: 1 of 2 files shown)


anayeaye:
@alukach Is there way to bust the cache? 60 seconds seems totally reasonable, but I can imagine a time in the future that we may want a private endpoint that we might hit when a new collection is added for a tenant. Discussion only, not change request

alukach:
> @alukach Is there way to bust the cache? 60 seconds seems totally reasonable, but I can imagine a time in the future that we may want a private endpoint that we might hit when a new collection is added for a tenant. Discussion only, not change request

There is no way to bust the cache, but being that the cache is in-memory and this is all run in a Lambda, I'm not sure if we need more than "wait a minute".  I see this as just a basic backstop, perhaps a more intelligent distributed caching could be useful someday to minimize lambda latency and excessive cold starts caused by cascading lambda executions

anayeaye (approved):
Approve but maybe we can dispatch this branch to test to observe this in the wild? @botanical

PR #547, opened 2025-10-22
build: speed up raster and stac lambda builds

ividito:
### Why?

`cdk synth` and `cdk diff` take a long time (5-15 minutes, depending on local resources and cache hits) because in order to interact with the CDK stack, the `app.synth()` call must build all of the docker assets we use. By changing how we bundle Docker assets, we can dramatically improve synth (and possibly deploy) times.

### What?

- @hrodmn made this work for [titiler-cmr](https://github.com/developmentseed/titiler-cmr/pull/81), and it carries over to our lambdas really well
- Dockerfiles are refactored, and produce small, optimized images that automatically go into ECR
- Ingest API excluded, because we use the same image with different entrypoints, and I haven't worked out how to integrate that yet
- DB bootstrapper is excluded because it's structured and executed differently.

Eventually we could go even further, by managing the ECR images ourselves. This would allow us to update a specific service without deploying the full stack (which would definitely accelerate deployments)

### Testing?

`time cdk synth` before (on `develop`)
Pasted image 20251016214746

`time cdk synth` after
Pasted image 20251016214132

Deployment to mcp-test and uah-sit (not done yet)
- [x] uah (to test runtimes)
- [ ] mcp (to test ECR deployment and permissions)

### DIFF
--- a/raster_api/infrastructure/construct.py
+++ b/raster_api/infrastructure/construct.py
@@ -13,6 +13,7 @@ from aws_cdk import (
     aws_lambda,
     aws_logs,
 )
+from aws_cdk.aws_ecr_assets import Platform
 from constructs import Construct
 
 from .config import veda_raster_settings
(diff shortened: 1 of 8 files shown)


raster_api/infrastructure/construct.py
```diff
@@ -37,17 +38,15 @@ def __init__(
         # TODO config
         stack_name = Stack.of(self).stack_name
 
-        veda_raster_function = aws_lambda.Function(
```
hrodmn, line 40: FYI there might be a performance hit on Lambda init time by switching from the zip package to a DockerImageFunction. We were forced into this change in titiler-cmr because we were bumping into the 250 MB limit for the function's code (many heavy dependencies). You should be able to use some of the same patterns from the Dockerfile to improve the build time for a typical `Function`, though.

Issue #521, opened 2025-08-25
[DS 25.4 Multi-Tenancy] Add list tenant endpoint


botanical:
Main Objective: https://github.com/NASA-IMPACT/veda-architecture/issues/630

Add tenant list endpoint. 
This ticket assumes that all parent tenant groups already exist in the realm. 

### Technical Details
See https://github.com/NASA-IMPACT/veda-architecture/pull/648 for more context.

### Acceptance Criteria
- [ ] Pre-req: https://github.com/NASA-IMPACT/veda-backend/issues/520 must be complete and tenant management schema and table must be deployed to dev
- [ ] Tenant endpoints are created/implemented with proper HTTP status codes

botanical:
Upon consideration, we might not want to make this endpoint. Enumerating all the tenants available might be a security concern. Currently, a user is able to check via `/auth/me` to see what groups they are a part of which I think is sufficient.

botanical:
Can use this ticket to capture work to create endpoint for Ingest UI to list writeable tenants a user has

PR #555, opened 2025-12-09
feat: add list tenants with create/update scopes endpoint

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/521

### What?/ Why?

Keycloak Client & Tenant Access Endpoint
- `KeycloakPDPClient` which contains functions:
  - `get_rpt` to request the requesting party token from keycloak
  - `check_permission` which checks to see is a user has a permission granted for a resource and scope
  - `get_tenants_with_create_update_access` which gets a list of tenants the user has create and update access to
  - base64 padding helper function (this is needed because the decode function requires proper padding or it will raise an error)
  - JWT permission extraction functions
- `/auth/tenants/writeable` endpoint added to Ingest API
- Ingest API config updated to include resource server client ID and secret env vars

### Testing?

-updated SIT envs to have `VEDA_KEYCLOAK_UMA_RESOURCE_SERVER_CLIENT_SECRET_NAME` and `VEDA_KEYCLOAK_SECRET_KMS_KEY_ARN`
- Deployed to sit https://sit.openveda.cloud/api/ingest/docs#/Auth/get_writable_tenant_access_auth_tenants_writable_get


stephenkilbourn (approved):

smohiudd:
@botanical I'm not seeing the `/auth/tenants/writeable` endpoint in the dev ingest api docs

ingest_api/infrastructure/config.py
```diff
@@ -90,6 +90,14 @@ class IngestorConfig(BaseSettings):
     keycloak_ingest_api_client_id: str = Field(description="Auth client ID")
 
     openid_configuration_url: AnyHttpUrl = Field(description="OpenID config url")
+
+    resource_server_client_id: Optional[str] = Field(
```
smohiudd: The pattern we've been using is to get the client id and secret from the secret manager. See Sm2a example with ingest api: https://github.com/NASA-IMPACT/veda-data-airflow/blob/dev/dags/veda_data_pipeline/utils/submit_stac.py#L49

So we could supply the secret arn instead.

ingest_api/runtime/src/main.py
```diff
@@ -215,6 +217,130 @@ def who_am_i(claims=Depends(oidc_auth.valid_token_dependency)):
...
+                access_token=user_access_token,
+                resource_type="item",
+            )
+
+            all_tenants = sorted(list(set(collection_tenants + item_tenants)))
+
+            return schemas.TenantAccessResponse(tenants=all_tenants)
+        finally:
+            # Clean up client
+            pdp_client.close()
```
sandrahoang686: ⛏️ : Do we need the nested trys? It looks like we are catching exceptions at the first level anyways? I think it also might be considered a good pattern to place the `finally` block after the exceptions block as its always going to run regardless of exceptions. ([exception handling](https://learnbyexample.github.io/100_page_python_intro/exception-handling.html#finally))

botanical: That's a really good point, and thanks for sharing the exception handling link @sandrahoang686 🙇 https://github.com/NASA-IMPACT/veda-backend/pull/555/changes/78600b760477cebe9e2acd8ecdf97909fe80eb1e I made some changes based on your feedback

smohiudd:
This looks good! Also a reminder we need to remove the old coginto auth from `/auth/veda_auth/main.py`. I'll make a ticket for this.

smohiudd (approved):

Issue #497, opened 2025-07-21
Upgrade to stac-fastapi v6 when released


anayeaye:
## What [WIP]
This issue anticipates the breaking changes that we will need to address in the stac-api app when v6 is released soon
https://github.com/stac-utils/stac-fastapi/compare/5.0.2..6.0.0

PR #561, opened 2026-01-05
fix: upgrade stac-fastapi to v6.1.5

ividito:
### Issue

#497 

### What?

- Update PG env var names
- Add flag to create writer connection pool for transactions
- Add a bunch of new tests for transactions, which also makes local OIDC work better

There were inconsistencies between how unit tests were running and how the actual service was behaving with the transactions changes. I had a really hard time getting meaningful errors out of the unit tests, so I converted them into integration tests that run against the `docker compose` services and make use of the existing auth integrations we have there. This was way more promising and has me feeling confident about the changes, but I'm still not sure what's wrong with the unit tests.

### Testing?

- ~Tests pass~ Unit tests fail 🤔, manual + integration tests pass
- Manually tested transactions, since the new default behaviour is to enable transactions with a new `stac-fastapi` environment variable. `veda-backend` already has a variable to enable the extension, which duplicates this functionality. We can continue to use this approach without breaking existing instances, or we can migrate to the alternative `stac-fastapi` variable.

### DIFF
--- a/stac_api/runtime/src/config.py
+++ b/stac_api/runtime/src/config.py
@@ -116,12 +116,11 @@ class _ApiSettings(Settings):
         if self.pgstac_secret_arn:
             secret = get_secret_dict(self.pgstac_secret_arn)
             return PostgresSettings(
-                postgres_host_reader=secret["host"],
-                postgres_host_writer=secret["host"],
-                postgres_dbname=secret["dbname"],
-                postgres_user=secret["username"],
-                postgres_pass=secret["password"],
-                postgres_port=secret["port"],
+                pghost=secret["host"],
+                pgdatabase=secret["dbname"],
+                pguser=secret["username"],
+                pgpassword=secret["password"],
+                pgport=secret["port"],
             )
         return PostgresSettings()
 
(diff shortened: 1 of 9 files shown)


stac_api/runtime/src/config.py
```diff
@@ -116,12 +120,11 @@ def postgres_settings(self) -> PostgresSettings:
...
-                postgres_host_writer=secret["host"],
-                postgres_dbname=secret["dbname"],
-                postgres_user=secret["username"],
-                postgres_pass=secret["password"],
-                postgres_port=secret["port"],
+                pghost=secret["host"],
+                pgdatabase=secret["dbname"],
+                pguser=secret["username"],
+                pgpassword=secret["password"],
+                pgport=secret["port"],
```
botanical, line 123: Should these be left as they were? https://github.com/stac-utils/stac-fastapi-pgstac/blob/6.1.5/stac_fastapi/pgstac/config.py I might be fully understanding but I noticed this when I pulled this branch locally and got a pydantic validation errors.

smohiudd: @botanical I got the same errors but the pgstac changes mention to [rename](https://github.com/stac-utils/stac-fastapi-pgstac/blob/main/CHANGES.md#changed-3)

ividito:
Unit tests fixed (thanks @smohiudd ), integration test changes moved over to #563

smohiudd (approved):

Issue #556, opened 2025-12-11
STAC API - Resource Extractor & Utilities (Foundation for PEP Middleware)


botanical:
Main Objective: https://github.com/NASA-IMPACT/veda-architecture/issues/686

Create resource extractor utilities to identify tenant ownership of collections and extract tenant info from requests. Implement Keycloak connection settings for authorization.

Something like:

- `extract_stac_resource_id(request)` - extracts collection/item ID from URL
- `extract_tenant_from_collection(collection)` - extracts tenant from collection properties

### Acceptance Criteria

- [x] STAC API can extract tenant from collections and requests
- [x] Keycloak client relevant env vars added to environment configuration
- [x] All new functions have unit tests and documentation

Issue #560, opened 2025-12-11
Implement Tenant Authorization in Ingest API Route Handlers (Phase III)


botanical:
Main Objective: https://github.com/NASA-IMPACT/veda-architecture/issues/686

Add tenant authorization checks to the ingest API's `POST` collections and `DELETE` collections endpoints. 

### Acceptance Criteria

- [ ] Discussion documented with team regarding delete endpoints
- [ ] Resource extractors functions expanded to include ingest API
- [ ] tenant authorization checks added to endpoints in Ingest API
- [ ] Ingest API has fallback scope-based authorization implemented

botanical:
Will close for now, we can reopen later if needed

PR #566, opened 2026-01-23
feat: create resource extractors for RPT (requesting party token) 

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/556

### What?/Why?

This PR  adds resource extractor functions that parse HTTP requests to extract resource ids and scopes that are needed for Keycloak's RPT endpoint.

In order to create a permission ticket or request an RPT from keycloak, we need
- resource id (needs to follow convention defined in our [keycloak config ](https://github.com/NASA-IMPACT/veda-keycloak/blob/main/keycloak-config-cli/config/dev/veda.yaml#L334))
- scope (action being performed)

Additional Context, from https://www.keycloak.org/docs/latest/authorization_services/index.html#_service_obtaining_permissions
```
Example of an authorization request when a client is seeking access to two resources protected by a resource server.

curl -X POST \
  http://${host}:${port}/realms/${realm-name}/protocol/openid-connect/token \
  -H "Authorization: Bearer ${access_token}" \
  --data "grant_type=urn:ietf:params:oauth:grant-type:uma-ticket" \
  --data "audience={resource_server_client_id}" \
  --data "permission=Resource A#Scope A" \
  --data "permission=Resource B#Scope B"
```

### Testing?

- Unit tests
- I will add integration tests when we create the actual policy enforcement point middleware that will use these functions 

You can also test this yourself trying different permission permutations. A successful request looks like:
```
curl -X POST \
  https://[HOST]/realms/veda/protocol/openid-connect/token \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -d 'grant_type=urn:ietf:params:oauth:grant-type:uma-ticket' \
  -d 'audience=uma-resource-server' \
  -d 'client_id=uma-resource-server' \
  -d 'client_secret=[redacted]' \
  -d 'permission=stac:collection:faketenant1:*#create' \
  -H "Authorization: Bearer TOKEN"
{"upgraded":false,"access_token":"redacted","token_type":"Bearer","not-before-policy":0}%
```

And one where you are not authorized because you either aren't in a Tenancy group or don't have sufficient permissions (determined by your role) in a tenancy, will yield

```
{"error":"access_denied","error_description":"not_authorized"}%                   
```

### DIFF
--- a/common/auth/veda_auth/resource_extractors.py
+++ b/common/auth/veda_auth/resource_extractors.py
@@ -110,3 +110,19 @@ async def extract_stac_resource_id(request: Request) -> Optional[str]:
         return None
 
     return None
+
+
+async def extract_ingest_resource_id(request: Request) -> Optional[str]:
+    """Extract resource ID for Ingest API requests"""
+    path = request.url.path
+    method = request.method
+
+    if path.endswith("/collections") and method == "POST":
+        return await _extract_collection_resource_id_from_post_body(request)
+
+    match = re.match(r".*?/collections/([^/]+)$", path)
+    if match and method == "DELETE":
+        collection_id = match.group(1)
+        return f"collection:{collection_id}"
+
+    return None
--- a/common/auth/veda_auth/resource_extractors.py
+++ b/common/auth/veda_auth/resource_extractors.py
@@ -22,6 +22,25 @@ STAC_ITEM_PUBLIC = "stac:item:public:*"
 STAC_COLLECTION_TEMPLATE = "stac:collection:{}:*"
 STAC_ITEM_TEMPLATE = "stac:item:{}:*"
 
+_COLLECTIONS_PATH_PATTERN = re.compile(r".*?/collections/([^/]+)$")
+_COLLECTIONS_ITEM_PATH_PATTERN = re.compile(r".*?/collections/([^/]+)/items/([^/]+)$")
+_COLLECTIONS_ITEMS_PATH_PATTERN = re.compile(r".*?/collections/([^/]+)/items$")
+_COLLECTIONS_BULK_ITEMS_PATH_PATTERN = re.compile(
+    r".*?/collections/([^/]+)/bulk_items$"
+)
+
+
+def _stac_collection_resource_id(request: Request) -> str:
+    """Return tenant-based or public STAC collection resource ID."""
+    tenant = getattr(request.state, "tenant", None)
+    return STAC_COLLECTION_TEMPLATE.format(tenant) if tenant else STAC_COLLECTION_PUBLIC
+
+
+def _stac_item_resource_id(request: Request) -> str:
+    """Return tenant-based or public STAC item resource ID."""
+    tenant = getattr(request.state, "tenant", None)
+    return STAC_ITEM_TEMPLATE.format(tenant) if tenant else STAC_ITEM_PUBLIC
+
 
 def _extract_tenant_from_body(
     body_data: Dict[str, Any], tenant_field: Optional[str] = None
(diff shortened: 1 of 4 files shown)


common/auth/veda_auth/resource_extractors.py
```diff
@@ -0,0 +1,128 @@
...
+    request: Request,
+) -> Optional[str]:
+    """Extract collection resource ID from POST/PUT collections request body"""
+    try:
+        request_body = await request.body()
+        if not request_body:
+            logger.warning(
+                "Cannot extract resource ID: empty body for collection operation"
+            )
+            return None
```
sandrahoang686: could raise an error here instead and log with except? so it follows the warning error flow.

```
async def _extract_collection_resource_id_from_post_body(
    request: Request,
) -> Optional[str]:
    """Extract collection resource ID from POST/PUT collections request body"""
    try:
        request_body = await request.body()
        if not request_body:
            raise HTTPException(
                status_code=status.HTTP_400_BAD_REQUEST,
                detail="Cannot extract resource ID: empty body for collection operation"
            )

        body_data = json.loads(request_body)
        tenant = _extract_tenant_from_body(body_data)
        if tenant:
            return f"stac:collection:{tenant}:*"
        else:
            return "stac:collection:public:*"
    except (json.JSONDecodeError, AttributeError, TypeError, HTTPException) as e:
        logger.warning(f"Failed to extract resource ID from collection body: {e}")
        return None
```

common/auth/veda_auth/resource_extractors.py
```diff
@@ -0,0 +1,128 @@
...
+
+        body_data = json.loads(request_body)
+        tenant = _extract_tenant_from_body(body_data)
+        if tenant:
+            return f"stac:collection:{tenant}:*"
+        else:
+            return "stac:collection:public:*"
```
sandrahoang686: ⛏️ : Noticing that `stac:collection:public:*` & `stac:item:public:*`are used in multiple places and at times multiple files - maybe we can make it into a constant so these strings are more centralized?

common/auth/veda_auth/resource_extractors.py
```diff
@@ -0,0 +1,128 @@
...
+    if match:
+        tenant = getattr(request.state, "tenant", None)
+        if tenant:
+            return f"stac:collection:{tenant}:*"
+        return "stac:collection:public:*"
+
+    if "/queryables" in path or "/search" in path:
+        return None
+
+    return None
```
sandrahoang686, line 114: ⛏️ : Since this is only returning either `stac:collection` or `stac:item` of either `{tenant}` or `public`, how about pulling out the conditional tenant check to return (either or) string to a fn?

```
async def extract_stac_resource_id(request: Request) -> Optional[str]:
    """Extract resource ID for STAC API requests
    Resource ID format matches Keycloak resource definitions (wildcard patterns):
    - Collections: "stac:collection:{tenant}:*" or "stac:collection:public:*"
    - Items: "stac:item:{tenant}:*" or "stac:item:public:*"
    """
    path = request.url.path
    method = request.method

    collections_path_pattern = re.compile(r".*?/collections/([^/]+)$")
    collections_list_items_path_pattern = re.compile(r".*?/collections/([^/]+)/items/([^/]+)$")
    collections_items_path_pattern = re.compile(r".*?/collections/([^/]+)/items$")
    collections_bulk_items_path_pattern = re.compile(r".*?/collections/([^/]+)/bulk_items$")

    def _collection_tenant_or_public(request: Request) -> str:
        """Return tenant-based STAC resource prefix or public fallback."""
        tenant = getattr(request.state, "tenant", None)
        return f"stac:collection:{tenant}:*" if tenant else "stac:collection:public:*"
    
    def _item_tenant_or_public(request: Request) -> str:
        tenant = getattr(request.state, "tenant", None)
        return f"stac:item:{tenant}:*" if tenant else "stac:item:public:*"

    if collections_path_pattern.match(path):
        if method in ("PUT", "PATCH"):
            return await _extract_collection_resource_id_from_post_body(request)

        return _collection_tenant_or_public(request)
    
    if collections_list_items_path_pattern.match(path):
        return _item_tenant_or_public(request)
    
    if collections_items_path_pattern.match(path) or collections_bulk_items_path_pattern.match(path):
        return _collection_tenant_or_public(request)

    if "/queryables" in path or "/search" in path:
        return None

    return None
```
could pull out the fns too for even more granular testing.

sandrahoang686 (commented):
🚀 my comments are just at the code level and unblocking. Someone with a better understanding of keycloak is probably more ideal to also review!

I wanted to test the CURL post as well but I was unsure where to access the client secret.

common/auth/veda_auth/resource_extractors.py
botanical: [9bbb583](https://github.com/NASA-IMPACT/veda-backend/pull/566/commits/9bbb5835b7be82754fa1fd2bc1c6ec2b9408135c)

anayeaye:
@botanical do you mind adding the new tests to the pull request action in github? https://github.com/NASA-IMPACT/veda-backend/blob/develop/.github/workflows/pr.yml#L80

botanical:
> @botanical do you mind adding the new tests to the pull request action in github? https://github.com/NASA-IMPACT/veda-backend/blob/develop/.github/workflows/pr.yml#L80

thanks for catching that! https://github.com/NASA-IMPACT/veda-backend/pull/566/commits/e80c79f1f4f288715eeb9bc403ca632b4fd5ff13 should be added now @anayeaye

common/auth/veda_auth/resource_extractors.py
```diff
@@ -0,0 +1,132 @@
...
+    """Extract tenant from request body JSON data"""
+    if tenant_field is None:
+        tenant_field = TENANT_FIELD
+
+    try:
+        tenant = body_data.get(tenant_field)
+        if tenant:
+            return tenant
+
+        properties = body_data.get("properties", {})
```
anayeaye: I am confused about looking for the tenant field in a properties object, don't we only define tenant as a top level field of a collection? For example: https://eic-staging.staging.earth.gov/api/stac/ghgc/collections. Also collections [do not have a properties object](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md) so we may not want to include this at all here?

botanical: Oh that's true, sorry that is confusing. For some reason I thought we might want this as a fallback. Let me remove it

anayeaye (approved):

common/auth/veda_auth/resource_extractors.py
```diff
@@ -0,0 +1,130 @@
...
+    ) or _COLLECTIONS_BULK_ITEMS_PATH_PATTERN.match(path):
+        return _stac_collection_resource_id(request)
+
+    if "/queryables" in path or "/search" in path:
+        return None
+
+    return None
+
+
+async def extract_ingest_resource_id(request: Request) -> Optional[str]:
```
anayeaye, line 117: Comment for discussion: I think `api/ingest/ingestions` will need to be accounted for in the resource extractor eventually as well when we begin enforcing tenant level auth in airflow, too? `/ingestions` is used by airflow for bulk item inserts

botanical: Yes, I agree with this! https://github.com/NASA-IMPACT/veda-backend/issues/560 I updated this to reflect that @anayeaye

Issue #558, opened 2025-12-11
Hybrid Authorization Routing for STAC API Write Operations (Phase II)


botanical:
Main Objective: https://github.com/NASA-IMPACT/veda-architecture/issues/686

Implement middleware routing logic that directs write operations to either UMA authorization (tenant collections) or stac-auth-proxy (tenantless collections), without full UMA implementation.

### Acceptance Criteria

- [x] Middleware intercepts write operations before handlers
- [x] Tenant collections route to UMA path
- [x] Tenantless collections pass through to stac-auth-proxy
- [x] Public read endpoints are not intercepted
- [x] Routing logic is testable and logged

Issue #557, opened 2025-12-11
PEP Middleware Proof of Concept


botanical:
Main Objective: https://github.com/NASA-IMPACT/veda-architecture/issues/686

## Summary

Implement a PEP middleware proof of concept that intercepts HTTP requests, extracts resources, and checks permissions using the Keycloak PDP client. This is a Proof of Concept ticket so it just needs to work for at least one endpoint.

## Requirements

- Create PEP middleware that intercepts HTTP requests before STAC API endpoints
- Test with a single endpoint (e.g., POST /collections)
- Extract resource identifiers (collection/item IDs) and tenant info from requests
- Integrate with Keycloak PDP client to check user permissions for extracted resources
- Return 403 Forbidden when permissions are denied, allow approved requests to proceed
- Add logging for permission checks and decisions

## Acceptance Criteria

- [ ] Middleware intercepts HTTP requests before endpoint handlers
- [ ] Can extract resources and tenant info from requests
- [ ] Successfully checks permissions via Keycloak PDP client
- [ ] Returns 403 for denied requests, allows approved requests to proceed
- [ ] Proof of concept test validates full flow for endpoint POST /collections

PR #569, opened 2026-02-14
feat(stac): policy enforcement point proof of concept

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/557
https://github.com/NASA-IMPACT/veda-backend/issues/558

### What?

- Creates proof of concept Policy Enforcement Point on STAC and Ingest API's POST /collections endpoints
- Handles writes for tenantless collections for users who are developers but do not belong to any tenant group
 
### Testing?

- Integration tests and ad hoc testing matrix (Tested on SIT https://sit.openveda.cloud/api/stac/docs)

| API | Test Case | Expected Result | Actual Result |
|--|--|--|--|
| STAC | User is an Admin of Tenant 1 | Can POST to tenant1 | ✅ |
| STAC | User is an Editor of Tenant 1 | Can POST to tenant1 | ✅ |
| STAC | User is NOT an Admin of Tenant 1, and NOT an editor | Can't POST to tenant1 | ✅ |
| Ingest | User is an Admin of Tenant 1 | Can POST to tenant1 | ✅  |
| Ingest | User is an Editor of Tenant 1 | Can POST to tenant1 | ✅  |
| Ingest | User is NOT an Admin of Tenant 1, and NOT an editor | Can't POST to tenant1 | ✅  |
| STAC | User is an Admin of Tenant 2 | Can POST to tenant2 | ✅ |
| STAC | User is an Editor of Tenant 2 | Can POST to tenant2 | ✅ |
| STAC | User is NOT an Admin of Tenant 2, and NOT an editor of Tenant 2 | Can't POST to tenant2 | ✅ |
| Ingest | User is an Admin of Tenant 2 | Can POST to tenant2 | ✅  |
| Ingest | User is an Editor of Tenant 2 | Can POST to tenant2 | ✅  |
| Ingest | User is NOT an Admin of Tenant 2, and NOT an editor of Tenant 2 | Can't POST to tenant2 | ✅  |
| STAC | User is an Admin of Tenant 3 | Can POST to tenant3 | ✅ |
| STAC | User is an Editor of Tenant 3 | Can POST to tenant3 | ✅ |
| STAC | User is NOT an Admin of Tenant 3, and NOT an editor of Tenant 3 | Can't POST to tenant3 | ✅ |
| Ingest | User is an Admin of Tenant 3 | Can POST to tenant3 | ✅  |
| Ingest | User is an Editor of Tenant 3 | Can POST to tenant3 | ✅  |
| Ingest | User is NOT an Admin of Tenant 3, and NOT an editor of Tenant 3 | Can't POST to tenant3 | ✅  |
| STAC | User is NOT a part of any tenancy but is a developer | Can post a public "tenant-less" collection |  ✅ |
| Ingest | User is NOT a part of any tenancy but is a developer | Can post a public "tenant-less" collection | ✅  |


common/auth/veda_auth/keycloak_client.py
```diff
@@ -7,14 +7,50 @@
...
 
+class TokenError(Exception):
+    """Raised when the access token is expired, revoked, or invalid.
+    Keycloak returns HTTP 401
+    """
+
+    def __init__(self, detail: str = "Access token is expired or invalid"):
+        """To use when there is a token error for RPT call"""
+        self.detail = detail
+        super().__init__(detail)
```
sandrahoang686, line 26:
```suggestion
    def __init__(self, message: str = "Access token is expired or invalid", code: int = 401):
        """To use when there is a token error for RPT call"""
        self.code = code
        super().__init__(message)
```
keep error `code` and use inherited `message` attribute?

common/auth/veda_auth/keycloak_client.py
```diff
@@ -7,14 +7,50 @@
...
+
+
+def parse_keycloak_from_openid_url(
+    openid_configuration_url: Union[str, Any]
+) -> Tuple[str, str]:
+    """Extract Keycloak base URL and realm from an OpenID discovery URL such as https://<host>/realms/<realm>/.well-known/openid-configuration"""
+    url_str = str(openid_configuration_url).strip() if openid_configuration_url else ""
+    if not url_str:
+        raise ValueError("Missing or empty OpenID configuration URL")
```
sandrahoang686:
```suggestion
    if not openid_configuration_url:
        raise ValueError("Missing or empty OpenID configuration URL")
    url_str = str(openid_configuration_url).strip()
```

since `openid_configuration_url` is req and if not passed in, should generate a `TypeError` we can probably move this `if` check to before and just need to eval the param, that way we error right away before any other logic and `str()` should always return some type of string unless if passed in nothing, it would still generate `""` ??

common/auth/veda_auth/keycloak_client.py
```diff
@@ -7,14 +7,50 @@
...
+
+    parsed = urlparse(url_str)
+    path = (parsed.path or "").rstrip("/")
+
+    if "/realms/" not in path:
+        raise ValueError(
+            "OpenID configuration URL must contain /realms/<realm>/ "
+            "(e.g. .../realms/my-realm/.well-known/openid-configuration)"
+        )
```
sandrahoang686, line 69: might be helpful to pass in `path` value to error for better debugging?

common/auth/veda_auth/pep_middleware.py
```diff
@@ -0,0 +1,183 @@
...
+            logger.exception(
+                "PEP: Keycloak check failed for resource_id=%s scope=%s: %s",
+                resource_id,
+                scope,
+                e,
+            )
+            return JSONResponse(
+                status_code=502,
+                content={"detail": "Authorization service temporarily unavailable"},
+            )
```
sandrahoang686, line 194: 🤔 error handling looks to be a bit redundant, not related or blocking to this PR at all but we could probably handle errors more cleaner and better with custom error handling logic that stays close to the source and bubbles up with a generic handler throughout veda-backend. Just a random thought i've noticed 💭 . Might create a ticket on this or work on this in spare time..

common/auth/veda_auth/keycloak_client.py
botanical: Hmm I am not seeing where the message attribute is inherited 😯 @sandrahoang686 I think detail works since it's consistent with this https://fastapi.tiangolo.com/tutorial/handling-errors/

sandrahoang686: sorry i should have linked => https://docs.python.org/3/library/exceptions.html#BaseExceptionGroup.message. It looks like `detail` could just be changed to `message` and then could be init'ed like so with inheritance 🤔 not entirely sure. We can keep like this too, dont have a strong opinion here.

sandrahoang686: maybe not https://fastapi.tiangolo.com/reference/exceptions/#fastapi.HTTPException.detail

common/auth/veda_auth/pep_middleware.py
botanical: https://github.com/NASA-IMPACT/veda-backend/pull/569/commits/5c0b8c1cfa06f071c46b121d4fa10e104d532c30

common/auth/veda_auth/resource_extractors.py
smohiudd, line 121: I should have asked this in the resource extractor PR, sorry...do we need to consider the ingest api `/ingestions/*` endpoints POST, PATCH and DELETE?

botanical: No worries @smohiudd. Yes, eventually to cover the ingestions endpoints but not all of them. We need to cover POST, but I think PATCH only updates the status and message right?

common/auth/veda_auth/keycloak_client.py
```diff
@@ -157,13 +206,69 @@ def _extract_permissions_from_jwt(self, jwt_token: str) -> List[Dict[str, Any]]:
...
+        permissions: List[Dict[str, Any]],
+        resource_id: str,
+        scope: str,
+    ) -> bool:
+        """Return True if permissions contain a grant for resource_id and scope
+
+        See https://www.keycloak.org/docs/latest/authorization_services/#_service_rpt_overview
+        """
+        for permission in permissions:
+            rsname = permission.get("rsname") or permission.get("resource_id")
```
ividito, line 244: I'm unclear on why `rsid` isn't relevant to check anymore

common/auth/veda_auth/keycloak_client.py
```diff
@@ -157,13 +206,69 @@ def _extract_permissions_from_jwt(self, jwt_token: str) -> List[Dict[str, Any]]:
...
+        Raises TokenError for 401, ResourceNotFoundError for 400 invalid_resource.
+        Re-raises unhandled status codes.
+        """
+        if error.response.status_code == 401:
+            logger.warning("Token rejected (401): %s", error.response.text)
+            raise TokenError(
+                "Access token is expired or invalid. Please re-authenticate."
+            ) from error
+        if error.response.status_code == 403:
+            return False
```
ividito: What do you think about raising another error (PermissionDeniedError?) here and catching it in the middleware? I think the `return False/raise Error` pattern is a bit confusing, this way we could replace https://github.com/NASA-IMPACT/veda-backend/pull/569/changes#diff-4857746bfd90942dbb37adfab0448bb3a8c3683230e3fadf002e8e89b615445cR187-R202 with another `except` block

stac_api/runtime/src/app.py
```diff
@@ -16,6 +16,7 @@
 )
 from src.extension import TiTilerExtension
 from stac_auth_proxy import configure_app
+from veda_auth.pep_middleware import PEPMiddleware
```
ividito: nit: should we put this import in the same conditional that instantiates it?

stac_api/runtime/src/app.py
```diff
@@ -142,6 +143,57 @@ async def lifespan(app: FastAPI):
...
+    from veda_auth.resource_extractors import extract_stac_resource_id
+
+    app.add_middleware(
+        PEPMiddleware,
+        pdp_client=_get_keycloak_pdp_client,
+        resource_extractor=extract_stac_resource_id,
+    )
+else:
+    logger.info(
+        "PEP middleware disabled,  openid_url=%s, secret_name=%s",
```
ividito:
```suggestion
        "PEP middleware disabled, openid_url=%s, secret_name=%s",
```

ividito (approved):
A couple questions and very minor suggestions. It all works as intended AFAIK. Ran most tests locally and also tried a few requests to SIT.

common/auth/veda_auth/keycloak_client.py
botanical: Yeah that's a good question! Basically we're not using `rsid` because  the RPT puts the name in rsname (e.g. "stac:item:faketenant2:*") and the UUID in rsid (and sometimes in resource_id). You can check this yourself by requesting an RPT and introspecting it but basically it will show something like
``` 
{
            "scopes": [
                "read"
            ],
            "rsid": "some-uuid",
            "rsname": "stac:item:faketenant2:*",
            "resource_id": "some-uuid",
            "resource_scopes": [
                "read"
            ]
        }
```

We need the name to match and to parse type/tenant, so we use rsname (and resource_id when it’s the name). rsid only ever gives us the UUID which we don't use. I guess really we don't need resource_id but I had that as a fallback because I wasn't sure where keycloak would put the resource name.

common/auth/veda_auth/keycloak_client.py
botanical: I think your suggestion makes sense! The bool/error return is confusing, sorry about that. 
[2176ff9](https://github.com/NASA-IMPACT/veda-backend/pull/569/commits/2176ff953c0c5c38273134ce9c315118cae496cf) What do you think of something like this?

common/auth/veda_auth/pep_middleware.py
```diff
@@ -0,0 +1,203 @@
...
+            request.url.path,
+        )
+
+        try:
+            pdp_client.check_permission(
+                access_token=token,
+                resource_id=resource_id,
+                scope=scope,
+            )
+        except TokenError as e:
```
smohiudd, line 150: Appreciate the structured and comprehensive error responses!

Issue #559, opened 2025-12-11
UMA Authorization Implementation for Tenant Collections (Phase II)


botanical:
Main Objective: https://github.com/NASA-IMPACT/veda-architecture/issues/686

Complete PEP Middleware for STAC API write operations and replace placeholder UMA routing with full Keycloak PDP authorization checks for tenant collections in STAC API write operations.

### Acceptance Criteria

- [x] UMA authorization checks work for tenant collections
- [x] Permission denials return 403 with clear error messages
- [x] Permission grants allow requests to proceed
- [x] Keycloak errors are handled gracefully
- [x] Full authorization flow is tested with real tokens

PR #573, opened 2026-03-13
feat: retrieve tenant from collection for items and ingest endpoints

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/559
And to begin supporting https://github.com/NASA-IMPACT/veda-backend/issues/560

### What?

- This handles item operations, making a look up on the collection object's tenant property to determine tenancy and then calling out to the PDP to ensure the user has permission to make that item request. 
- Extends PEP to ingest DELETE endpoint

### Testing?

https://sit.openveda.cloud/api/ingest/docs
https://sit.openveda.cloud/api/stac/docs

|API | Test Case | Expected Result | Actual Result |
|--|--|--|--|
|STAC | User in tenant1 is able to update an item in tenant1 | Allow | ✅   |
|STAC | User in tenant1 is able to delete an item in tenant1  | Allow |  ✅ |
|STAC | User in tenant2 attempt to delete item in tenant1 | Deny |  ✅ |
|STAC | User in tenant1 attempts to patch an item in tenant2| Deny |  ✅ |
|STAC | User with no tenant memberships updates a public item | Allow |  ✅ |
| Ingest | User is an Admin of Tenant 1 and tries to delete a tenant1 collection | Allow |✅ |
| Ingest | User is an Editor of Tenant 1  and tries to  delete a collection in tenant1 | Deny | ✅  |
| Ingest | User is an Editor of Tenant 2 and tries to delete a collection in tenant1 | Deny |  ✅ |
| Ingest | User is an Admin of Tenant 2 and tries to delete a collection in tenant1 | Deny |✅ |

### DIFF
--- a/common/auth/veda_auth/resource_extractors.py
+++ b/common/auth/veda_auth/resource_extractors.py
@@ -118,15 +118,10 @@ async def _extract_collection_resource_id_from_post_body(
         return None
 
 
-async def extract_stac_resource_id(request: Request) -> Optional[str]:
-    """Extract resource ID for STAC API requests
-    Resource ID format matches Keycloak resource definitions (wildcard patterns):
-    - Collections: STAC_COLLECTION_TEMPLATE or STAC_COLLECTION_PUBLIC
-    - Items: STAC_ITEM_TEMPLATE or STAC_ITEM_PUBLIC
-    """
-    path = request.url.path
-    method = request.method
-
+async def _extract_collection_stac_resource_id(
+    request: Request, path: str, method: str
+) -> Optional[str]:
+    """Extract resource ID for collection endpoints, or None if not a collection path"""
     if _COLLECTIONS_CREATE_PATH_PATTERN.match(path) and method == "POST":
         return await _extract_collection_resource_id_from_post_body(request)
 
@@ -135,6 +130,13 @@ async def extract_stac_resource_id(request: Request) -> Optional[str]:
             return await _extract_collection_resource_id_from_post_body(request)
         return _stac_collection_resource_id(request)
 
+    return None
+
+
+async def _extract_item_stac_resource_id(
+    request: Request, path: str, method: str
+) -> Optional[str]:
+    """Extract resource ID for item endpoints, or None if not an item path"""
     if _COLLECTIONS_ITEM_PATH_PATTERN.match(path):
         # For single item operations, prefer collection tenant when available
         match = _COLLECTIONS_ITEM_PATH_PATTERN.match(path)
@@ -163,6 +165,27 @@ async def extract_stac_resource_id(request: Request) -> Optional[str]:
             if tenant:
                 return STAC_ITEM_TEMPLATE.format(tenant)
         return _stac_collection_resource_id(request)
+
+    return None
+
+
+async def extract_stac_resource_id(request: Request) -> Optional[str]:
+    """Extract resource ID for STAC API requests
+
+    Resource ID format matches Keycloak resource definitions (wildcard patterns):
+    - Collections: STAC_COLLECTION_TEMPLATE or STAC_COLLECTION_PUBLIC
+    - Items: STAC_ITEM_TEMPLATE or STAC_ITEM_PUBLIC
+    """
+    path = request.url.path
+    method = request.method
+
+    collection_id = await _extract_collection_stac_resource_id(request, path, method)
+    if collection_id is not None:
+        return collection_id
+
+    item_id = await _extract_item_stac_resource_id(request, path, method)
+    if item_id is not None:
+        return item_id
 
     if "/queryables" in path or "/search" in path:
         return None
--- a/common/auth/veda_auth/resource_extractors.py
+++ b/common/auth/veda_auth/resource_extractors.py
@@ -10,7 +10,7 @@ import json
 import logging
 import os
 import re
-from typing import Any, Dict, Optional
+from typing import Any, Awaitable, Callable, Dict, Optional
 
 from fastapi import HTTPException, Request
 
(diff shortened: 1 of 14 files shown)


smohiudd (approved):
`collection_tenant_resolver` and querying the db looks like a good way to get the tenant for items endpoints. No proposed changes from me.

ingest_api/runtime/src/collection_publisher.py
```diff
@@ -30,6 +35,64 @@ def delete(self, collection_id: str):
...
+        """Return tenant field from collection JSON in PgSTAC, or None if not found"""
+        tenant_field = settings.tenant_filter_field
+        logger.info(
+            "Resolving tenant for collection %s using tenant_field=%s",
+            collection_id,
+            tenant_field,
+        )
+        creds = get_db_credentials(os.environ["DB_SECRET_ARN"])
+        try:
+            with PgstacDB(dsn=creds.dsn_string, debug=True) as db:
```
botanical, line 43: I initially tried to use the collection loader / VEDALoader to get information on the collection, but because the tenancy field is non standard, it didn't yield any results compared to directly querying like so (lines 49-50).

common/auth/veda_auth/resource_extractors.py
```diff
@@ -87,31 +133,82 @@ async def _extract_collection_resource_id_from_post_body(
...
+    match = _COLLECTIONS_PATH_PATTERN.match(path)
+    if match:
         if method in ("PUT", "PATCH"):
             return await _extract_collection_resource_id_from_post_body(request)
+        if method == "DELETE":
+            collection_id = match.group(1)
+            tenant = await _collection_tenant_for_item(request, collection_id)
+            if tenant:
+                return STAC_COLLECTION_TEMPLATE.format(tenant)
+            return _stac_collection_resource_id(request)
```
sandrahoang686, line 152: Can probably remove this? Since without this, it will next land on line 153 which is returning same?

common/auth/veda_auth/resource_extractors.py
```diff
@@ -87,31 +133,82 @@ async def _extract_collection_resource_id_from_post_body(
...
         return _stac_collection_resource_id(request)
 
+    return None
+
+
+async def _extract_item_stac_resource_id(
+    request: Request, path: str, method: str
```
sandrahoang686: is `method` being used?

common/auth/veda_auth/resource_extractors.py
```diff
@@ -87,31 +133,82 @@ async def _extract_collection_resource_id_from_post_body(
...
+        return _stac_collection_resource_id(request)
+
+    if _COLLECTIONS_BULK_ITEMS_PATH_PATTERN.match(path):
+        match = _COLLECTIONS_BULK_ITEMS_PATH_PATTERN.match(path)
+        collection_id = match.group(1) if match else None
+        if collection_id:
+            tenant = await _collection_tenant_for_item(request, collection_id)
+            if tenant:
+                return STAC_ITEM_TEMPLATE.format(tenant)
         return _stac_collection_resource_id(request)
```
sandrahoang686: Looks like the condition logic for items and bulk items and very similar, maybe it can be combined to simplify?

something like
```
      for pattern in (_COLLECTIONS_ITEMS_PATH_PATTERN, _COLLECTIONS_BULK_ITEMS_PATH_PATTERN):
          if match := pattern.match(path):
              tenant = await _collection_tenant_for_item(request, match.group(1))
              if tenant:
                  return STAC_ITEM_TEMPLATE.format(tenant)
              return _stac_collection_resource_id(request)
``` 
?

common/auth/veda_auth/resource_extractors.py
botanical: I think we need it because, otherwise, it wouldn't land on line 153 but it would land on the outermost return which would be `None`, and we want to make sure it falls back to get the collection resource ID using the request if it doesn't manage to resolve using the resolver (looking up in the DB in the case of DELETEs)

common/auth/veda_auth/resource_extractors.py
botanical: 😯 good call out! Let me remove it

common/auth/veda_auth/resource_extractors.py
botanical: https://github.com/NASA-IMPACT/veda-backend/pull/573/commits/60b529e3ff426ff97a9c8de1c513c1be34297dbc

common/auth/veda_auth/resource_extractors.py
botanical: https://github.com/NASA-IMPACT/veda-backend/pull/573/commits/60b529e3ff426ff97a9c8de1c513c1be34297dbc

sandrahoang686 (approved):
code wise - lgtm!

sandrahoang686 (approved):

ingest_api/runtime/src/collection_publisher.py
```diff
@@ -30,6 +35,54 @@ def delete(self, collection_id: str):
...
 
+    def get_collection_tenant(self, collection_id: str) -> Optional[str]:
+        """Return tenant field from collection JSON in PgSTAC, or None if not found"""
+        tenant_field = settings.tenant_filter_field
+        creds = get_db_credentials(os.environ["DB_SECRET_ARN"])
+        try:
+            with PgstacDB(dsn=creds.dsn_string, debug=True) as db:
+                collection_content = db.query_one(
+                    "SELECT content FROM collections WHERE id=%s",
+                    (collection_id,),
```
botanical, line 46: ^ spicy, would love some eyes on this

PR #572, opened 2026-03-10
feat: update to generate map links and render preview assets for all renders config assets, unit test

sandrahoang686:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/567

**First AC** => **Thumbnail and map links generated for all COG assets in a collection's render configuration** ✅ 

---
**Changes in PR:**
* adds unit tests for the `_search_base` method in `stac_api/runtime/src/core.py` file
   * tests pass ✅ 
* updates `_search_base` method to generate Map Link Items for all render config assets (originally just dashboard)
* updates `_search_base` method to generate render preview assets for all render config assets (originally just dashboard)

---

**Current Test Outcome Generated:**
```
{
   "type":"FeatureCollection",
   "features":[
      {
         "id":"test-item",
         "type":"Feature",
         "collection":"test-collection",
         "links":[
            {
               "title":"Map of Item for ndvi",
               "href":"https://fake-titiler.example.com/collections/test-collection/items/test-item/WebMercatorQuad/map?nodata=0&assets=ndvi",
               "rel":"preview",
               "type":"text/html"
            },
            {
               "title":"Map of Item for colorIR",
               "href":"https://fake-titiler.example.com/collections/test-collection/items/test-item/WebMercatorQuad/map?nodata=0&assets=colorIR&bidx=1&bidx=2&bidx=3&rescale=0%2C255",
               "rel":"preview",
               "type":"text/html"
            },
            {
               "title":"Map of Item for dashboard",
               "href":"https://fake-titiler.example.com/collections/test-collection/items/test-item/WebMercatorQuad/map?nodata=-9999&assets=burnRatio&rescale=-1%2C1",
               "rel":"preview",
               "type":"text/html"
            }
         ],
         "assets":{
            "rendered_preview_ndvi":{
               "title":"Rendered preview for ndvi",
               "href":"https://fake-titiler.example.com/collections/test-collection/items/test-item/preview.png?nodata=0&assets=ndvi",
               "rel":"preview",
               "roles":[
                  "overview"
               ],
               "type":"image/png"
            },
            "rendered_preview_colorIR":{
               "title":"Rendered preview for colorIR",
               "href":"https://fake-titiler.example.com/collections/test-collection/items/test-item/preview.png?nodata=0&assets=colorIR&bidx=1&bidx=2&bidx=3&rescale=0%2C255",
               "rel":"preview",
               "roles":[
                  "overview"
               ],
               "type":"image/png"
            },
            "rendered_preview_dashboard":{
[...]

Issue #567, opened 2026-01-24
Expand renders support for multi-asset items and for additional services (WMTS, titiler-x, etc)


anayeaye:
## What
The veda-backend stac-api dynamicly generates thumbnail  assets and map preview links for the `dashboard` object in Collection metadata--extend the existing method to generate thumbnail and map preview links for all objects configured in a Collection's renders definition. _This issue can probably be separated into 3 or 4 separate deliverables._

### Generate links for all COG assets
The stac-api currently generates dynamic links when renders configuration is provided for a `dashboard` object ([stac-api/runtime/src/core.py](https://github.com/NASA-IMPACT/veda-backend/blob/1e8b19f2b9c241d214f7d3756150a685dc6baac9/stac_api/runtime/src/core.py#L55)). Extend this to generate dynamic assets and links for all objects in the renders configuration (found in Collections of items with multiple assets).

#### AC
- [ ] Thumbnail and map links generated for all COG assets in a collection's render configuration. _(Review Status)_

### Add support for alternate tilers in RenderConfig
Dynamic renders links currently use the default raster-api (TiTiler). Add an optional `tiler_url` parameter that can override the default raster-api url to support Collections with assets that use alternative tilers such as titiler-cmr and titiler-multidim.

#### AC
- [ ] A STAC collection containing assets that should be visualized using *titiler-cmr* demonstrates functional thumbnail and map preview links _(Review Status)_
- [ ] A STAC Collection containing assets that should be visualized using *titiler-multidim* demonstrates functional thumbnail and map preview links _(Review Status)_

### Generate WMTS thumbnails and previews for collections with web-map-links
When a Collection indexing ESRI server data contains a properly formatted web-map-links object, generate links with visualization parameters from the web-map-links object.

#### AC
- [ ] A STAC Collection that indexes data served via an ESRI service with web-map-links demonstrates functional pre-formatted WMTS links

### Generate WMTS links for use in ArcGIS
> Note https://github.com/NASA-IMPACT/veda-architecture/issues/695 should first be completed to determine the appropriate template for WMTS links that work directly in ArcGIS online
This deliverable focuses on exposing COG (and possibly titiler-x) data indexed in a VEDA STAC Collection in ArcGIS via WMTS links. A new `wmts` object will need to be defined to differentiate these dynamic links/assets.

#### AC
- [ ] Pre-configured WMTS urls expose COG (and possibly titiler-x) assets web map tiles for usage in ArcGIS

[...]

PR #574, opened 2026-03-17
feat: add optional tiler url override for renders config

sandrahoang686:
Related Ticket: https://github.com/NASA-IMPACT/veda-backend/issues/567

**BLOCKED** by https://github.com/NASA-IMPACT/veda-backend/pull/572 _(this is gonna need #572 to merge first and then will need to pull from this branch to resolve conflicts)_

### DIFF
--- a/stac_api/runtime/tests/test_core.py
+++ b/stac_api/runtime/tests/test_core.py
@@ -1,0 +1,93 @@
+"""
+Unit tests for VedaCrudClient._search_base in core.py
+"""
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from src.core import VedaCrudClient, ExtendedSearchRequest
+
+from stac_fastapi.pgstac.core import CoreCrudClient
+from stac_fastapi.pgstac.types.search import PgstacSearch
+
+
+def make_client():
+    """Create a VedaCrudClient instance without calling __init__."""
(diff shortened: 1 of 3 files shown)


stac_api/runtime/tests/test_core.py
```diff
@@ -0,0 +1,93 @@
...
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        """Initialize mocks"""
+        default_tiler = "https://fake-titiler.example.com"
+        override_tiler = "https://fake-titiler-override.example.com"
+        mock_data = {
+            "collections": ["noaa-emergency-response"],
+            "limit": 5,
+            "tiler_url": override_tiler,
+        }
```
sandrahoang686, line 33: @anayeaye This is how i'm currently going about the alternate tiler_url override. The request would look like...
```
curl -X POST http://localhost:8081/search \
    -H "Content-Type: application/json" \
    -d '{"collections": ["noaa-emergency-response"], "limit": 5, "tiler_url": "https://fake-titiler-override.example.com"}'
```
is this the correct way to go about this? opening a draft to see if i'm on the right path 👀

anayeaye: I haven't tried any of this but it reads right. The goal is to make it possible for someone to use renders with a titiler other than the one that is a part of the veda-backend deployment. Like titiler-cmr, or titiler-multidim

PR #576, opened 2026-03-24
fix: pin titiler and mangum

ividito:
### Issue

No issue - related [thread in Slack](https://developmentseed.slack.com/archives/C045K5LAFA9/p1774368650506379)

### What?

- Pin dependencies so we don't use `titiler>=2.0` or `mangum>0.19` for the raster API

### Testing?

- Tests pass locally

### DIFF
--- a/raster_api/runtime/setup.py
+++ b/raster_api/runtime/setup.py
@@ -11,9 +11,9 @@ inst_reqs = [
     "rio-tiler>=7.0,<8.0",
     "titiler.pgstac==1.5.0",
     # use highest available titiler based on rio-tiler and titiler-pgstac pins (not 2.0 yet though)
-    "titiler.core<0.20,<2.0",
-    "titiler.mosaic<2.0",
-    "titiler.extensions[cogeo]<2.0",
+    "titiler.core<0.20",
+    "titiler.mosaic<0.20",
+    "titiler.extensions[cogeo]<0.20",
     "starlette-cramjam>=0.3,<0.4",
     # based on AWS observability requirements
     "aws_xray_sdk>=2.6.0,<3",
--- a/raster_api/runtime/setup.py
+++ b/raster_api/runtime/setup.py
@@ -10,15 +10,16 @@ inst_reqs = [
     # highest-tested versions of raster_api dependencies
     "rio-tiler>=7.0,<8.0",
     "titiler.pgstac==1.5.0",
-    # use highest available titiler based on rio-tiler and titiler-pgstac pins
-    "titiler.core<0.20",
-    "titiler.mosaic",
-    "titiler.extensions[cogeo]",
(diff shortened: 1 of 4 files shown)


smohiudd (approved):

raster_api/runtime/setup.py
```diff
@@ -10,15 +10,16 @@
...
     "rio-tiler>=7.0,<8.0",
     "titiler.pgstac==1.5.0",
-    # use highest available titiler based on rio-tiler and titiler-pgstac pins
-    "titiler.core<0.20",
-    "titiler.mosaic",
-    "titiler.extensions[cogeo]",
+    # use highest available titiler based on rio-tiler and titiler-pgstac pins (not 2.0 yet though)
+    "titiler.core<0.20,<2.0",
+    "titiler.mosaic<2.0",
+    "titiler.extensions[cogeo]<2.0",
```
vincentsarago:
```suggestion
    "titiler.core==0.19.3",
    "titiler.mosaic==0.19.3",
    "titiler.extensions[cogeo]==0.19.3",
```
I would do this

raster_api/runtime/setup.py
```diff
@@ -10,15 +10,16 @@
...
     "starlette-cramjam>=0.3,<0.4",
     # based on AWS observability requirements
     "aws_xray_sdk>=2.6.0,<3",
     "aws-lambda-powertools>=1.18.0",
     "python-multipart==0.0.7",
+    "mangum<=0.19",
```
vincentsarago: better change https://github.com/NASA-IMPACT/veda-backend/blob/fix/broken-raster-deps/raster_api/runtime/Dockerfile#L8

vincentsarago (dismissed):

hrodmn (approved):

hrodmn:
It would be a good idea to update the API runtimes to use uv.lock files when installing dependencies so the Lambda runtime builds are more deterministic (the base Lambda image can still change). We did this in eoapi-cdk a while back to avoid dependency upgrade landmines like this! https://github.com/developmentseed/eoapi-cdk/pull/208

raster_api/runtime/setup.py
```diff
@@ -10,15 +10,16 @@
...
-    "titiler.mosaic",
-    "titiler.extensions[cogeo]",
+    "titiler.mosaic<0.20",
+    "titiler.extensions[cogeo]<0.20",
     "starlette-cramjam>=0.3,<0.4",
     # based on AWS observability requirements
     "aws_xray_sdk>=2.6.0,<3",
     "aws-lambda-powertools>=1.18.0",
     "python-multipart==0.0.7",
+    "mangum<=0.19",
```
vincentsarago: we don't need this

Issue #596, opened 2026-04-28
Document Mulitenancy Configuration and Operations for Instance Support


botanical:
### Description
Document how to configure and operate multitenancy so that instance operators can enable, disable, and troubleshoot independently.

### Acceptance Criteria

- [x] Configuration levers for enabling/disabling multitenancy (both filtering and authorization) documented
- [x] Operational guidance for common workflows and troubleshooting documented
- [ ] Documentation accessible from the wiki

PR #597, opened 2026-05-12
fix(stac-api): root path bug and docs update

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-backend/issues/596
https://github.com/NASA-IMPACT/veda-architecture/issues/754

### What?

- Updates the docs to include more information about disabling multitenancy features, distinguishing between filtering and authorization
- Updates TenantExtractionMiddleware logic for root paths with trailing slash (so it skips trailing slash removal for `{root_path}/`)
- Adds unit tests for TenantExtractionMiddleware

Tested tenant extraction change in UAH-SIT.
| Test setup | Result |
| --|--|
| Envs: STAC_AUTH_PROXY_ENABLED=False, used v14.0.0-rc.16 | redirect errors on https://sit.openveda.cloud/api/stac/ |
| Envs: STAC_AUTH_PROXY_ENABLED=False, used this branch's commit SHA | redirect errors resolved on https://sit.openveda.cloud/api/stac/| 

The reason we didn't notice this before is because we never had an instance tested where we had the Tenant Extraction middleware running while Tenant Filtering (AKA stac-auth-proxy) was disabled.

### DIFF
--- a/stac_api/runtime/README.md
+++ b/stac_api/runtime/README.md
@@ -4,7 +4,7 @@
 
 Tenant-related behavior comes from **several middleware components**. Each row names one component, what it does, and how it is turned on (some are always on; others depend on environment variables).
 
-| Component | Role | Enabled when: |
+| Component | Role | Enabled when |
 | --- | --- | --- |
 | **`TenantExtractionMiddleware`** | Parses tenant URLs such as `{root_path}/{tenant}/collections`, sets `request.state.tenant`, and rewrites `scope["path"]` so routing matches the usual STAC paths. | **Always** (see `app.py`). Independent of `VEDA_STAC_ENABLE_STAC_AUTH_PROXY`. |
 | **`TenantLinksMiddleware`** | For JSON STAC responses, rewrites `links` so clients stay on the tenant-prefixed URL when a tenant is present. | **Always** (see `app.py`). |
@@ -13,7 +13,7 @@ Tenant-related behavior comes from **several middleware components**. Each row n
 
 **How this fits together:**
 
-- **Tenant in the URL** (`/api/stac/veda/collections`, …) is handled by the **two URL-tenant middleware components** in the table so the core STAC app still sees `/api/stac/collections` while `request.state.tenant` is set. That works even when the auth proxy wrapper is **off**.
+- **Tenant in the URL** (`/api/stac/{Tenant=veda}/collections`, …) is handled by the **two URL-tenant middleware components** in the table so the core STAC app still sees `/api/stac/collections` while `request.state.tenant` is set. That works even when the auth proxy wrapper is **off**.
 - **Tenant in collection metadata** (filtering which collections/items appear for a tenant) is driven by the **filters registered through `stac-auth-proxy`** when the proxy is **on**. Turning the proxy off removes that integration; it does **not** remove URL parsing or link rewriting from the tenant middleware.
 - **Transactions** still require `VEDA_STAC_ENABLE_TRANSACTIONS=True` and, per configuration rules, `VEDA_STAC_ENABLE_STAC_AUTH_PROXY=True` when transactions are enabled.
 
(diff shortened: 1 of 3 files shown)


stac_api/runtime/README.md
```diff
@@ -1,6 +1,25 @@
...
+| **`TenantLinksMiddleware`** | For JSON STAC responses, rewrites `links` so clients stay on the tenant-prefixed URL when a tenant is present. | **Always** (see `app.py`). |
+| **`stac-auth-proxy` (`configure_app`)** | Wraps the STAC app: OpenAPI + OpenID Connect for clients, registers **`CollectionFilter` / `ItemFilter`** so list/search respect collection tenant metadata (`VEDA_TENANT_FILTER_FIELD`, default `eic:tenant`), and requires OAuth scopes on write routes. | When `VEDA_STAC_OPENID_CONFIGURATION_URL` is set **and** `VEDA_STAC_ENABLE_STAC_AUTH_PROXY=True`. |
+| **`PEPMiddleware`** | Keycloak UMA checks on configured write routes. | When `VEDA_STAC_OPENID_CONFIGURATION_URL` is set **and** `VEDA_KEYCLOAK_UMA_RESOURCE_SERVER_CLIENT_SECRET_NAME` is set. |
+
+**How this fits together:**
+
+- **Tenant in the URL** (`/api/stac/veda/collections`, …) is handled by the **two URL-tenant middleware components** in the table so the core STAC app still sees `/api/stac/collections` while `request.state.tenant` is set. That works even when the auth proxy wrapper is **off**.
```
anayeaye: Can we make it more clear that in this example the tenant is 'veda'? Maybe Tenant=veda in the URL would work?

stac_api/runtime/README.md
```diff
@@ -23,16 +42,37 @@ To enable a multi-tenant STAC backend with transaction support and authenticatio
...
 
 **Field Format:**
 
 The `eic:tenant` field should contain a string identifier for the tenant.
+
+## Disabling Multitenancy
+
+Multi-tenancy has several parts. The **two optional layers** below (auth-proxy filter integration and PEP) are toggled independently. Both require `VEDA_STAC_OPENID_CONFIGURATION_URL` to be set, plus each layer’s own second flag.
+
+| Layer | What it does | Enabled when |
```
anayeaye: I know we are technically covered, but maybe we could do a Disabled when column? Just to be extra clear for this section on disabling multitenancy

stac_api/runtime/README.md
botanical: Like this https://github.com/NASA-IMPACT/veda-backend/pull/597/changes/6066ed0ba6a6d279da70709b717c989131820e6b#diff-d9358104ba7e3320761121d9bc6ec25ca836f666dc09788bb7eb0c147b444f9eR16 ?

stac_api/runtime/README.md
botanical: https://github.com/NASA-IMPACT/veda-backend/pull/597/changes/6066ed0ba6a6d279da70709b717c989131820e6b#diff-d9358104ba7e3320761121d9bc6ec25ca836f666dc09788bb7eb0c147b444f9eR63 Added a new `Disabled when` column. Let me know your thoughts @anayeaye

stac_api/runtime/README.md
```diff
@@ -23,16 +42,37 @@ To enable a multi-tenant STAC backend with transaction support and authenticatio
...
 The `eic:tenant` field should contain a string identifier for the tenant.
+
+## Disabling Multitenancy
+
+Multi-tenancy has several parts. The **two optional layers** below (auth-proxy filter integration and PEP) are toggled independently. Both require `VEDA_STAC_OPENID_CONFIGURATION_URL` to be set, plus each layer’s own second flag.
+
+| Layer | What it does | Enabled when | Disabled when |
+| --- | --- | --- | --- |
+| **Tenant-scoped filtering** | Wraps the app with `stac-auth-proxy`: OpenID Connect, registers `CollectionFilter` / `ItemFilter` for tenant-aware list/search, OAuth-scoped write endpoints | `VEDA_STAC_OPENID_CONFIGURATION_URL` is set **and** `VEDA_STAC_ENABLE_STAC_AUTH_PROXY=True` | `VEDA_STAC_OPENID_CONFIGURATION_URL` is not set **or** `VEDA_STAC_ENABLE_STAC_AUTH_PROXY=False` |
+| **PEP authorization** | Adds Keycloak UMA middleware that checks per-tenant resource permissions on protected routes | `VEDA_STAC_OPENID_CONFIGURATION_URL` is set **and** `VEDA_KEYCLOAK_UMA_RESOURCE_SERVER_CLIENT_SECRET_NAME` is set | `VEDA_STAC_OPENID_CONFIGURATION_URL` is not set **or** `VEDA_KEYCLOAK_UMA_RESOURCE_SERVER_CLIENT_SECRET_NAME` is not set |
```
anayeaye:
```suggestion
| **Tenant-scoped filtering** | Wraps the app with `stac-auth-proxy`: OpenID Connect, registers `CollectionFilter` / `ItemFilter` for tenant-aware list/search, OAuth-scoped write endpoints | `VEDA_STAC_OPENID_CONFIGURATION_URL` is set **and** `VEDA_STAC_ENABLE_STAC_AUTH_PROXY=True` | `VEDA_STAC_ENABLE_STAC_AUTH_PROXY=False` **or** UNSET |
| **PEP authorization** | Adds Keycloak UMA middleware that checks per-tenant resource permissions on protected routes | `VEDA_STAC_OPENID_CONFIGURATION_URL` is set **and** `VEDA_KEYCLOAK_UMA_RESOURCE_SERVER_CLIENT_SECRET_NAME` is set |  `VEDA_KEYCLOAK_UMA_RESOURCE_SERVER_CLIENT_SECRET_NAME` is not set |
```

Not sure if I got the table editing right but I think `VEDA_STAC_OPENID_CONFIGURATION_URL` could be set in cases where multitenacy is NOT enabled (as in a single tenant backend that allows authorized stac api transactions). So I tried to remove that condition from both of the disabled when conditions. Also I think enable stac auth proxy defaults to False?

Hmm I am actually a little confused because it looks like stac auth proxy has to be true for any transactions (even for backends without multenancy? or will transactions not even work for backends without multitenancy?) https://github.com/NASA-IMPACT/veda-backend/blob/24766f30667a9ae21ad0770b6adf8248dedc2e69/stac_api/infrastructure/config.py#L71

botanical: > Hmm I am actually a little confused because it looks like stac auth proxy has to be true for any transactions (even for backends without multenancy? or will transactions not even work for backends without multitenancy?)


Yes, and it's validated here https://github.com/NASA-IMPACT/veda-backend/blob/24766f30667a9ae21ad0770b6adf8248dedc2e69/stac_api/infrastructure/config.py#L85-L103 to check that, in order to turn on transactions on the STAC API, `enable_stac_auth_proxy` must be True (as well as some other checks)

> Not sure if I got the table editing right but I think VEDA_STAC_OPENID_CONFIGURATION_URL could be set in cases where multitenacy is NOT enabled (as in a single tenant backend that allows authorized stac api transactions). So I tried to remove that condition from both of the disabled when conditions. Also I think enable stac auth proxy defaults to False?


This is accurate. `VEDA_STAC_OPENID_CONFIGURATION_URL` could be set and isn't tied to multitenancy but I think it's more technically accurate that unsetting `VEDA_STAC_OPENID_CONFIGURATION_URL` and `VEDA_STAC_ENABLE_STAC_AUTH_PROXY` disables filtering. Though I see your point that we probably won't use `VEDA_STAC_OPENID_CONFIGURATION_URL` to disable tenant filtering.

anayeaye (approved):
Approve but can we fix the PR conventional commit to reflect that we are also delivering a stac fast api root path bug fix?

PR #605, opened 2026-05-27
fix(stac): upgrade stac auth proxy to v1.1.1

botanical:
### Issue

https://github.com/NASA-IMPACT/veda-architecture/issues/762

### What?

- Upgrade stac-auth-proxy to v1.1.1
- Adds secondary defense by using `request.scope["path"]`  as recommended by https://www.secwest.net/starlette

> Upgrade Starlette to 1.0.1 or later. Rebuild and redeploy every container, virtualenv, and bundled artifact that pins or vendors Starlette. Bundled installs are common in LLM tooling; pip list on the host is not enough. Audit images.

> Secondary (defense in depth)
Replace request.url and request.url.path with request.scope["path"] in every middleware, dependency, and decorator that makes security decisions. Grep the codebase. This bug class will recur; reading the un-reconstructed value is the durable fix.

### Testing

- Tested authorization on STAC API
- Tested endpoints `POST /collection`, `DELETE /collection`
- Verified that tenant extraction and addition to links works 
- Verified that tenant filtering works https://dev.openveda.cloud/api/stac/faketenant2/collections
- Verified that STAC browser works as usual

### DIFF
--- a/common/auth/veda_auth/resource_extractors.py
+++ b/common/auth/veda_auth/resource_extractors.py
@@ -35,6 +35,14 @@ _COLLECTIONS_ITEMS_PATH_PATTERN = re.compile(COLLECTIONS_ITEMS_PATH_RE)
 _COLLECTIONS_BULK_ITEMS_PATH_PATTERN = re.compile(COLLECTIONS_BULK_ITEMS_PATH_RE)
 
 
+def _request_path(request: Request) -> str:
+    """Use request.scope["path"] primarily"""
+    scope_path = request.scope.get("path")
+    if isinstance(scope_path, str) and scope_path:
+        return scope_path
+    return request.url.path
+
+
 def _stac_collection_resource_id(request: Request) -> str:
     """Return tenant-based or public STAC collection resource ID."""
     tenant = getattr(request.state, "tenant", None)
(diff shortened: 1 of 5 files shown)


common/auth/veda_auth/resource_extractors.py
```diff
@@ -35,6 +35,14 @@
 _COLLECTIONS_BULK_ITEMS_PATH_PATTERN = re.compile(COLLECTIONS_BULK_ITEMS_PATH_RE)
 
 
+def _request_path(request: Request) -> str:
+    """Use request.scope["path"] primarily"""
+    scope_path = request.scope.get("path")
+    if isinstance(scope_path, str) and scope_path:
+        return scope_path
+    return request.url.path
```
ividito: `scope_path` will always be a `str` or `None`, and so `if scope_path` will evaluate the same as what we have here. I think instead of calling this function, it would be simpler to use:

```path = request.scope.get("path") or request.url.path```

or (in `common/auth/veda_auth/pep_middleware.py`)

```path = (request.scope.get("path") or request.url.path).rstrip("/") or "/"```

botanical: https://github.com/NASA-IMPACT/veda-backend/pull/605/commits/1478b1b47c1ae7d6d5f6f4a58b1af3459d316c80

ividito (approved):

anayeaye (approved):

Issue #589, opened 2026-04-03
Stac-browser should only be built on version change (security improvement)


anayeaye:
## What
We currently build the stac-browser using eoapi-cdk on every deployment regardless of changes the stac-browser version. This adds unnecessary build time and introduces more opportunities for vulnerable versions of dependencies to be installed (such as https://github.com/NASA-IMPACT/veda-architecture/issues/710#issuecomment-4184085464).

## AC
- [ ] Stac browser website is only built and deployed in CICD when the stac browser version environment variable is updated.

jjfrench:
I have a broader question: should the STAC Browser version be updated? Currently VEDA is on v3.2.0 in dev. v5.0.0 was released last week - can I just change this or does this go through a stakeholder?

PR #624, opened 2026-07-25
chore: uv support

jjfrench:
### Issue

I apologize to whoever is reviewing this.

- https://github.com/NASA-IMPACT/veda-architecture/issues/791

I will say, this is not a fix-all solution. This was purely to get uv support added and the workflows running. There are still plenty of things to address, such as dependency versions/pinning, pydocs, and refining Dockerfiles, but this PR is already absurdly massive.

*Most files with only a few changes are purely format changes

### What?

- Adds support for uv package management
- Moves linting to Ruff
- Addresses 117 mypy errors
- Patches local tests: adds a retry to requests
- Updates several ingest_api dependencies (not the latest version) due to cqi error: boto3, s3fs, fsspec

### Why?

- UV support, but Ruff, mypy, and running tests blocked the workflows

### Testing?

- https://github.com/NASA-IMPACT/veda-backend/actions/runs/30299500381

Things to look for:
- cdk-diff: number of stacks changed


jjfrench:
On the bright side, 7,625 of the 8,432 additions are just uv.lock files

.pre-commit-config.yaml
```diff
@@ -1,31 +1,23 @@
...
-    hooks:
-      - id: mypy
-        language_version: python
-        additional_dependencies: ['types-requests', 'types-cachetools', 'types-PyYAML']
+  # re-enable pydocstyle once we have docstrings in place
+  # - repo: https://github.com/PyCQA/pydocstyle
+  #   rev: 5.1.1
+  #   hooks:
+  #     - id: pydocstyle
+  #       language_version: python
```
jjfrench, line 23: Doesn't seem like this was being used, considering the number of errors it raised. If re-enabled, uncomment the Dependabot section to manage revisions.

Makefile
jjfrench, line 1: Taken from veda-routes. Trying to add some consistency across VEDA repositories

scripts/run-local-tests.sh
```diff
@@ -33,17 +48,24 @@ cleanup() {
...
 # Run tests
 echo "--- Running stac and raster tests ---"
-python -m pytest .github/workflows/tests/ -vv -s
+uv run pytest .github/workflows/tests/ -vv -s
 
 # Run ingest unit tests
 echo "--- Running ingest api runtime tests ---"
-NO_PYDANTIC_SSM_SETTINGS=1 python -m pytest --cov=ingest_api/runtime/src ingest_api/runtime/tests/ -vv -s
+# Must ping PGSTAC_VERSION in multiple places due to version management outside of repository
+PGSTAC_VERSION=0.9.6
```
jjfrench, line 58: The Dockerfile installs the version through a build argument in CDK, but since this is running using `uv run`, it was either to pin it here or in the pyproject.toml[test]

ividito (approved):

common/auth/veda_auth/resource_extractors.py
```diff
@@ -1,4 +1,4 @@
-"""" Resource Extractors to use in PEP Middleware.
+""" " Resource Extractors to use in PEP Middleware.
```
botanical:
```suggestion
"""Resource Extractors to use in PEP Middleware.
```

.github/workflows/pr.yml
```diff
@@ -67,24 +66,26 @@ jobs:
...
+          uv sync --project ingest_api/runtime --group test
 
       - name: Install reqs for stac api
-        run: python -m pip install stac_api/runtime/
+        run: uv sync --project stac_api/runtime --group test
 
       - name: Ingest unit tests
-        run: NO_PYDANTIC_SSM_SETTINGS=1 python -m pytest ingest_api/runtime/tests/ -vv -s
+        env:
+          PGSTAC_VERSION: 0.9.6
```
botanical, line 80: Could we get the version we need from our env instead?

.github/workflows/tests/test_raster.py
```diff
@@ -34,11 +38,32 @@ def setup(
...
+    def _get_with_retry(self, url, attempts=5, delay=0.5, **kwargs):
+        """Retry transient startup transport errors when local services are cold."""
+        for attempt in range(1, attempts + 1):
+            try:
+                return httpx.get(url, timeout=HTTP_TIMEOUT, **kwargs)
+            except (httpx.TimeoutException, httpx.TransportError):
+                if attempt == attempts:
+                    raise
+                time.sleep(delay)
+
```
botanical, line 60: nice ✨

Makefile
```diff
@@ -0,0 +1,36 @@
...
+	uv run ruff format
+
+diff:
+	uv run npx aws-cdk@2 diff -v
+
+STACKS = --all
+ARGS =
+
+deploy:
+	uv run npx aws-cdk@2 deploy --outputs-file ./cdk-outputs.json $(STACKS) $(ARGS)
```
botanical: I think we need to update where it outputs the file because of what we have here https://github.com/NASA-IMPACT/veda-deploy/blob/dev/.github/workflows/deploy.yml#L108-L121? 
We could also remove --outputs-file from the Makefile's deploy target so it stays generic, and pass it via ARGS in this action instead:
```
make deploy ARGS="--require-approval never --outputs-file  ${HOME}/cdk-outputs.json"
``` 
? open to whatever you think is best though @jjfrench !

.github/workflows/pr.yml
jjfrench: I'm not sure, which env?

Makefile
jjfrench: I like the idea of passing it as additional ARGS! Keeps the makefile simple

Do you know what `${Home}` gets set as?

.github/workflows/pr.yml
botanical: `VEDA_DB_PGSTAC_VERSION` 🤔 maybe

jjfrench: I'm trying to install this as a requirement to run the tests

I could use the same AWS Secrets step from actions/deploy to grab the version and inject it here, but I'm not sure if tests should be making requests to secrets

Makefile
botanical: I do not, sorry! I bet we could echo it though

common/auth/README.md
```diff
@@ -19,7 +19,7 @@ The `KeycloakPDPClient` enables applications to:
 The client is part of the `veda_auth` package. You can install it with:
 
 ```bash
-uv pip install common/auth/
+uv sync --project common/auth
```
jjfrench, line 22: @aliziel, the configuration I was trying to remember during the review was
> [tool.uv]
> package = false

and this is only in the root `pyproject.toml` - which overrides the default of `true` I believe.  Either way, the dependencies should still be installable

jjfrench:
Can someone smarter than me tell me why this is failing now? 🥲 I don't think that the last commit affected the predeploy step

ividito:
>Can someone smarter than me tell me why this is failing now? 🥲 I don't think that the last commit affected the predeploy step

I reproduced this locally and narrowed it down to an upstream `postcss` change. We could try limiting the npm install to `--before=2026-07-27` or upgrading stac-browser to 3.2.0+ (where package.lock was committed in stac-browser).

jjfrench:
@ividito Upgrading might not be a bad idea unless you have hard requirements for the stac-browser version. (v5.0.0 was merged today and should be released soon). Is this tag value configured in Secrets Manager? 

Could you lmk how you were able to run this locally?

ividito:
CDK eats the underlying error message, but eoapi-cdk basically pulls and builds from the given git tag, so I did this to start seeing errors:

```
git clone --depth 1 --branch v3.1.0 https://github.com/radiantearth/stac-browser.git sb
cd sb
npm install
npm run build -- --catalogUrl=https://dev.openveda.cloud/api/stac
```

Then I tried a known build that worked:

```
git clone --depth 1 --branch v3.1.0 https://github.com/radiantearth/stac-browser.git sb2
cd sb2
npm install --before=2026-03-25
npm run build -- --catalogUrl=https://dev.openveda.cloud/api/stac
```

Claude did the actual work of narrowing down the root cause, but you can check the work with this:

```
cd sb2
npm install --no-save postcss@8.5.25
npm run build -- --catalogUrl=https://dev.openveda.cloud/api/stac   # fails
```

>Is this tag value configured in Secrets Manager?

I don't think so - I'm pretty sure we're using the default in `config.py` and not overriding it.

jjfrench:
@ividito I'll bump it to the minimum working version and then VEDA can decide if they want to update to the latest v5.0.0 in the #589 

Thanks for taking the time and building those locally!

botanical (approved):

PR #635, opened 2026-08-12
fix(raster): de-duplicate vendored PROJ database

ividito:
Slack convo: https://developmentseed.slack.com/archives/C045K5LAFA9/p1786567144333719

Our raster API image is too big! This trims 9-10MB, giving us 7MB of headroom. I don't think this breaks anything - `proj` gets pointed to the DB from `rasterio`, which is newer and likely better.


anayeaye (approved):
I tested out a variety of raster api requests including statistics and clipping ot a geometry. Works as expected.

PR #649, opened 2026-09-10
fix: cap RDS proxy connections so raster spikes queue instead of overloading Postgres

### LINKED ISSUE #646: pgstac statement_timeout and admission control database configuration
From the Disasters workshop post-mortem: 
- Kills the 3,505 orphaned searches per hour, the amplifier
- Stops 900 Lambdas from pushing 845 connections onto two vCPUs; queues at the proxy instead of thrashing CPU; trades saturation for fast, honest throttles

### AC
- [ ] statement_timeout set on pgstac role (just under 30s Lambda timeout)
- [ ] RDS proxy MaxConnectionsPercent reduced from 100 to something like 80
- [ ] Lambda reserved concurrency configured (~100-150)

CarsonDavis:
During the disasters dashboard user testing on 2026-09-09, the raster API was down for about 33 minutes. The load was modest, around 38 tile requests a second, but enough of them missed the cache at once that the 2-vCPU dev database went into a spiral it could not climb out of while requests kept coming. Load testing on dev reproduced the failure and showed the trigger is how many database connections are open at once, not the request rate. This PR caps that number at the RDS Proxy, so the excess waits in line at the proxy instead of piling onto Postgres.

## Background

Every raster request runs in its own Lambda invocation, and each Lambda instance opens exactly one database connection. Between the Lambdas and Postgres sits an RDS Proxy. It has a setting, `MaxConnectionsPercent`, that caps how many connections the proxy may open to the database, expressed as a percentage of the database's own `max_connections`. AWS defaults it to 100%.

The catch is what `max_connections` measures. RDS derives it from the instance's memory (`DBInstanceClassMemory / 9531392`), so on a 16 GiB db.r5.large it is about 1,700. But that instance has only two vCPUs, and CPU is what runs out. Two cores can run a handful of queries at full speed. Add more concurrent queries and every one of them slows down, so the same throughput costs more CPU, responses take longer, Lambdas stay alive longer, more Lambdas open more connections, and eventually everything breaks.

So the setting we want is not "100 percent of what memory allows" but "a few concurrent queries per vCPU". On a db.r5.large that is about 10 connections, or 1 percent. Requests past the cap wait in the proxy's queue, where waiting is cheap, instead of inside Postgres, where waiting makes everything else slower too.

The proxy has a second setting, `MaxIdleConnectionsPercent`, for how many connections it keeps open once they go unused. AWS defaults it to 50 and requires it to be no greater than the max. At 50 the proxy held the 672 connections the incident-reproduction run opened for two hours after load stopped. This PR sets both to the same value.

## What this changes

- **New setting `VEDA_DB_PROXY_MAX_CONNECTIONS_PERCENT`, default 1.** It is passed to the proxy as both the max and the idle percent. It only matters when `VEDA_DB_USE_RDS_PROXY` is on.
- **The cap is shared by every client of the proxy.** The raster API, the STAC API, the ingest API and the bulk ingestor all connect through the same proxy as the same role, so ten connections is the budget for all of them. Ingest holds one of those connections, pinned, for the length of a load, which leaves raster nine for those minutes. Under a raster storm, an ingest or STAC query waits its turn at the proxy, a few seconds at the loads tested, and then runs at full speed; before the cap the same storm gave it a database answering in hundreds of milliseconds per query and a 30 s timeout.
- **What the cap does not protect is the Lambda pool.** Past about 200 req/s of uncached tiles, raster fills the account's 900 concurrent Lambda instances, and from then on ingest and STAC invocations are throttled at the door the same as excess tiles. That was already true during the incident. The fix is reserved concurrency on the ingest and STAC functions so raster cannot take their slots, and it belongs in a follow-up.

## What the tests showed

All runs are uncached tile requests against the dev stack on a db.r5.large, reading the second minute of each run.

**Reproducing the incident at the AWS default (100 / 50).** A step from idle to 90 req/s spawned about 438 Lambda instances at once. Connections went from 149 (the proxy's retained idle pool) to 421 in a minute and to 662 at the peak of the ramp that followed. CPU pegged at 99 percent, queries that usually take 3 ms took 720 ms, Lambda p50 hit 11 s and p99 hit the 30 s timeout. CPU recovered within minutes of load stopping, but the connections settled at 672 and were still there two hours later.

**Applying a cap drains the pool.** Setting the proxy to 3 / 3 took those 672 connections down to 46 within a minute. Under any cap the connection count never moves with load again: it sits flat at the cap through every run below, including the ones that overload Lambda.

**Choosing the number.** The first cap tried was 3 percent (46 connections), which held a step to 90 req/s with no queueing. We then ran the same 150 req/s at four caps:

| Cap | DB connections | RDS CPU | Query inside Postgres | Lambda p50 | Failures |
| --- | --- | --- | --- | --- | --- |
| 5 % | 80 | 85 % | 16 ms | 9.3 s | 19 % |
| 3 % | 46 | 99 % | 14 ms | 5.4 s | ~0 |
| 2 % | 30 | 77 % | 7.9 ms | 160 ms | 0 |
| 1 % | 12 | 66 % | 3.7 ms | 71 ms | 0 |

Two things to read carefully in that table. The 5 percent run started on a fleet that had been idle for an hour, so its second minute served about a fifth fewer requests than the others; its failure rate is not comparable and its CPU is lower than it would be at equal work, which is why 3 percent reads hotter than 5. And the 3 percent row's "~0 failures" hides a run that had every one of the account's 900 Lambda instances alive and a 5.4 s median response. Below 3 percent the picture is clean: at 12 connections each query runs at its idle-time speed, while at 46 the queries contend and take four times longer for the same throughput. 

**Testing 1 percent to failure.** With the cap at 1 percent (12 connections on RDS) we pushed the load up until something broke:

| Load | Duration | Requests | Lambda instances | RDS CPU | Query inside Postgres | Lambda p50 | Wait at the proxy | Failures |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 150 req/s | 2 min | 18,000 | 147 | 66 % | 3.7 ms | 71 ms | 0.15 s | 0 |
| 150 req/s | 10 min soak | 90,000 | 101 to 250 | 54 to 67 % | 3.5 to 4.7 ms | 59 to 166 ms | up to 0.26 s | 0 |
| 200 req/s (198 delivered) | 2 min | 23,776 | 798 | 69 to 96 % | 5.0 to 5.4 ms | 0.75 to 2.8 s | 1.2 to 2.4 s | 0 |
| 300 req/s (278 delivered) | aborted at 1 min 32 s | 25,593 | 900 (the account limit) | 68 to 88 % | 3.6 to 4.6 ms | 6.5 s | 3.7 s | 38 %, all throttles |

Nothing broke at the database. At 300 req/s the failures are Lambda throttles: each invocation spends 3 to 4 s waiting for one of the twelve connections, so at that rate the fleet would need about 2,000 instances and the account has 900. Excess tiles fail fast instead of hanging, the database never leaves 12 connections and 5 ms queries, and CPU is back to idle within two minutes of load stopping. The no-error ceiling for uncached tiles is about 200 req/s. For scale, the 2026-09-09 session peaked at 39 req/s at the edge.

## Why 1 percent

We also ran the 300 req/s break test at 2 and 3 percent (the 3 percent run started on a partly cold fleet, so its served count is a little low):

| Cap | Served per minute | Rejected | RDS CPU | Query inside Postgres |
| --- | --- | --- | --- | --- |
| 3 % | 7,360 | 59 % | 99 % | 10 to 11.5 ms |
| 2 % | 9,037 | 41 % | 99 % | 10 to 12.5 ms |
| 1 % | 10,013 | 38 % | 88 % | 4.6 ms |

1 percent wins every column at both loads. At 150 req/s, 5 and 3 percent fill the whole Lambda pool and push median responses past 5 s; 2 percent is fine on failures but each query costs twice the CPU and the median is twice as long. At 300 req/s, 2 percent serves fewer tiles than 1 percent and pegs the CPU doing it.

The reason is where the overload lands. At 2 percent and above enough queries reach Postgres to contend for the two cores, so per-query cost climbs and the database is what saturates, which is the failure we are fixing. At 1 percent the twelve connections, not the CPU, are the ceiling: the excess sits in the proxy queue and the Lambda pool, both of which clear the moment load stops. Ten connections is about five concurrent queries per vCPU, which is the rule the sweep confirms.


.example.env
```diff
@@ -9,6 +9,8 @@ VEDA_DB_SCHEMA_VERSION=0.1.0
 VEDA_DB_SNAPSHOT_ID=[OPTIONAL BUT **REQUIRED** FOR ALL DEPLOYMENTS AFTER BASING DEPLOYMENT ON SNAPSHOT]
 VEDA_DB_PUBLICLY_ACCESSIBLE=TRUE
 VEDA_DB_USE_RDS_PROXY=[OPTIONAL]
+VEDA_DB_PROXY_MAX_CONNECTIONS_PERCENT=[OPTIONAL, DEFAULTS TO 80]
+VEDA_DB_STATEMENT_TIMEOUT=[OPTIONAL, DEFAULTS TO 28s]
```
amarouane-ABDELHAK: Can we make this optional without any default value? I'm concerned that if ingest takes more than n seconds, we'll have to redeploy repeatedly, bumping the timeout each time, to get it right . It would be better to remove the condition and redeploy,  run the ingest, then deploy again with a reasonable value.

amarouane-ABDELHAK: Also shouldn't the statement_timeout to be forced to be less than the API gateway timeout < 30s

.example.env
```diff
@@ -9,6 +9,8 @@ VEDA_DB_SCHEMA_VERSION=0.1.0
 VEDA_DB_SNAPSHOT_ID=[OPTIONAL BUT **REQUIRED** FOR ALL DEPLOYMENTS AFTER BASING DEPLOYMENT ON SNAPSHOT]
 VEDA_DB_PUBLICLY_ACCESSIBLE=TRUE
 VEDA_DB_USE_RDS_PROXY=[OPTIONAL]
+VEDA_DB_PROXY_MAX_CONNECTIONS_PERCENT=[OPTIONAL, DEFAULTS TO 80]
```
amarouane-ABDELHAK: The default value in AWS for this is 100. If our default is lower than that but doesn't actually bound load, I don't think it buys us anything. Can we at least match the AWS default, or make this fully optional with no default value (falling back to the AWS default)?

bitner:
OK, so first line of defense. Do NOT use the default RDS setting for MAX_CONNECTIONS. Always set that number to be a realistic value of how many concurrent connections you expect to get. Don't try to be too optimistic here either, make sure that it is a realistic value. THIS is the postgres setting that should then drive what you set for WORK_MEM and the general rule of thumb is that MAX_CONNECTIONS * WORK_MEM should be approximately or a bit less than the value that is set for SHARED_BUFFERS.

It is FAR FAR FAR better to return an error fast to a user from running out of connections than to do something that would end up with OOM memories, processes that end up hanging, etc. While you are at it, make sure that you have good settings for STATEMENT_TIMEOUT and IDLE_IN_TRANSACTION_TIMEOUT and LOCK_TIMEOUT. Note that you can set these separately on the role that you use to connect to the database from the API with (which is hopefully not the admin role, but a separate login role that inherits from pgstac_read). A role used for ingest, would want these limits (and possibly WORK_MEM and MAINTENANCE_WORK_MEM) to be higher than for the read role.

I do not have direct experience with RDS Proxy, but in general for a connection pooler, you can set the number of max connections there to be a higher value than the max connections on the database itself as the pooler will then just wait to be able to get a connection when one becomes available from the database. Your true line of defense is to get your database settings dialed in first. Once you have those in a good place, then start looking at the proxy settings.

database/infrastructure/config.py
```diff
@@ -60,6 +60,17 @@ class vedaDBSettings(BaseSettings):
...
+        1,
+        description=(
+            "Percent of the instance's max_connections the RDS proxy may open; "
+            "excess clients queue at the proxy. Keep it small: max_connections "
+            "scales with instance memory, but the real limit is a few concurrent "
+            "queries per vCPU."
+        ),
+        ge=1,
+        le=100,
+    )
```
anayeaye, line 73:
```suggestion
    proxy_max_connections_percent: int = Field(
        100,
        description=(
            "Percent of the instance's max_connections the RDS proxy may open; "
            "excess clients queue at the proxy. Keep it small: max_connections "
            "scales with instance memory, but the real limit is a few concurrent "
            "queries per vCPU."
            "https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_rds/DatabaseProxyOptions.html"
        ),
        ge=1,
        le=100,
    )
    proxy_max_idle_connections_percent: int = Field(
        50,
        description=(
            "Controls how actively the proxy closes idle database connections "
            "in the pool. High values leave more idle connections open; low "
            "values close idle client connections and return the underlying "
            "database connections to the pool."
            "https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_rds/DatabaseProxyOptions.html"
        ),
        ge=1,
        le=100,
    )
```

database/infrastructure/construct.py
```diff
@@ -270,6 +270,8 @@ class DatabaseConfig(TypedDict):
                 role=proxy_role,
                 require_tls=False,
                 debug_logging=False,
+                max_connections_percent=veda_db_settings.proxy_max_connections_percent,
+                max_idle_connections_percent=veda_db_settings.proxy_max_connections_percent,
```
anayeaye, line 274:
```suggestion
                max_idle_connections_percent=veda_db_settings.proxy_max_idle_connections_percent,
```

.example.env
```diff
@@ -9,6 +9,7 @@ VEDA_DB_SCHEMA_VERSION=0.1.0
 VEDA_DB_SNAPSHOT_ID=[OPTIONAL BUT **REQUIRED** FOR ALL DEPLOYMENTS AFTER BASING DEPLOYMENT ON SNAPSHOT]
 VEDA_DB_PUBLICLY_ACCESSIBLE=TRUE
 VEDA_DB_USE_RDS_PROXY=[OPTIONAL]
+VEDA_DB_PROXY_MAX_CONNECTIONS_PERCENT=[OPTIONAL, DEFAULTS TO 1]
```
anayeaye, line 12:
```suggestion
VEDA_DB_PROXY_MAX_CONNECTIONS_PERCENT=[OPTIONAL, DEFAULTS TO 100]
VEDA_DB_PROXY_MAX_IDLE_CONNECTIONS_PERCENT=[OPTIONAL, DEFAULTS TO 50]
```

anayeaye (commented):
Can we keep the defaults and override in the environment variables instead? https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_rds/DatabaseProxyOptions.html

anayeaye:
I made a couple suggestions to keep defaults but I am now realizing that also implies updating the description a bit to explain the two new env variables and strongly recommend overriding both to 1% for an R5 instance with 8 GiB RAM: 1 vCPU.

anayeaye:
I am going to get these changes to maintain cdk defaults with the option to tune RDS proxy connection (and I will add the env variables to match what was originally proposed in this PR for the disasters instances).

I also added a backlog issue for us to get together with @bitner and improve the RDS parameters used. https://github.com/NASA-IMPACT/veda-architecture/issues/866

anayeaye (approved):

ividito (approved):

